PR #389, opened 2015-01-05
Adding explanation for when a message is public, moderated and confirmed

### LINKED ISSUE #387: What does it mean to have a "confirmed" message?
I accessed my instance and I saw the messages are "confirmed", but I don't know if that means that the sender confirmed their email address or that the reciever got the email. In any case, both pieces of information are valuable and there needs to be a description for this field (at least a tooltip).

lfalvarez:
this closes #387.

@martinszy are the explanations ok? if not could you help us with a bit of feedback?


tmtmtmtm:
"This message can not be seen by everyone" is a little ambiguous. Perhaps better to express it in terms of who *can* see it?

lfalvarez:
Cool! how about "This message can be seen by you only"

tmtmtmtm:
:+1:

Issue #441, opened 2015-01-14
Store raw incoming emails


tmtmtmtm:
Currently only the processed form is stored — make sure we store the original too, so we can reprocess it, handle errors etc.

PR #452, opened 2015-01-14
Store all raw email

lfalvarez:
it now saves every incoming email and relates it to an answer when it is possible


lfalvarez:
It is related to #452

lfalvarez:
This one closes #441

PR #454, opened 2015-01-15
removing psycopg2 from requirements

lfalvarez:
Removing psycopg2 from requirements


lfalvarez:
Ok I'm thinking that this is required so I'm closing this issue

PR #470, opened 2015-01-19
Flake8

duncanparkes:
Sorry this is a bit of a monster, and I'm not expecting anyone to read it all.

I've run flake8 over the codebase using the settings file that Django uses, and these are the changes I think need making to make it (almost) happy. On SayIt we have a git pre-commit hook which makes sure we produce PEP8 satisfying code - it might be nice to do the same thing here.

Most of the stuff in this commit is just for style reasons to make it more readable. I'll make another commit in a second with a couple more things which flake8 brought up which look more important to me.

I'll be interested to see what travis has to say about this as I have 5 test failures on my dev environment even with origin/master!

Cheers,

Duncan (hoping shortly to get on with some proper issues :-))


lfalvarez:
:smile: that's awesome!

duncanparkes:
Hi @lfalvarez 

It would be really good to get this merged now that it's passing as every time there's new stuff on master I have to rebase it!

Cheers,

Duncan

PS Try the git pre-commit hook thing :-)

lfalvarez:
You mean this? http://flake8.readthedocs.org/en/latest/vcs.html

duncanparkes:
Actually, I just put a fairly minimal pre-commit hook example in the root of the repo, which you can turn on by sym-linking it to the .git/hooks directory.

We use a pre-commit hook on sayit for compiling the sass as well, so it would have refused to install that hook. Having said that, I'm sure it's more sophisticated than what I added.

lfalvarez:
Thanks ;)

PR #475, opened 2015-01-20
Update django version

duncanparkes:
It should be fine to update to the latest security release of Django, I think.

I'm also trying removing a couple of lines which I don't think are used from travis.xml. We'll see what happens.


lfalvarez:
Cool I'm looking at it now =)

duncanparkes:
I'd let travis have a look first!

PR #476, opened 2015-01-20
WIP Intermittent 404 rather than 302

duncanparkes:
Don't merge this yet, I just want to see what travis has to say about something.

### DIFF
--- a/nuntium/user_section/tests/user_section_views_tests.py
+++ b/nuntium/user_section/tests/user_section_views_tests.py
@@ -1,6 +1,5 @@
 from global_test_case import GlobalTestCase as TestCase, popit_load_data
 from django.core.urlresolvers import reverse
-from django.core.urlresolvers import reverse as original_reverse
 from ...models import WriteItInstance
 from django.contrib.auth.models import User
 from django.test.client import Client, RequestFactory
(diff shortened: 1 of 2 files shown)


duncanparkes:
This fixes two test failures which happen consistently for me, but which don't occur on travis. The problem was that the view for editing an answer notification template or a confirmation template was using the pk of the writeitinstance to look up the template (by it's pk attribute) some way up the inheritance chain of get_object.

I'm not sure why this has not been failing on travis. Perhaps there is some peculiarity of the environment which causes those two primary keys to always be the same?

nuntium/user_section/tests/user_section_views_tests.py
```diff
@@ -1,23 +1,24 @@
-from global_test_case import GlobalTestCase as TestCase, popit_load_data
 from django.core.urlresolvers import reverse
```
lfalvarez: Cool!

nuntium/user_section/views.py
```diff
@@ -185,26 +184,30 @@ def dispatch(self, *args, **kwargs):
...
             raise Http404
+
+    def get_object(self):
+        self.check_ownership()
         self.object = super(WriteItInstanceOwnerMixin, self).get_object()
         return self.object
 
 
-class UpdateTemplateWithWriteitMixin(UpdateView, LoginRequiredMixin, WriteItInstanceOwnerMixin):
+class WriteItRelatedModelMixin(SingleObjectMixin):
```
lfalvarez: This is very helpful with the naming

nuntium/user_section/views.py
```diff
@@ -185,26 +184,30 @@ def dispatch(self, *args, **kwargs):
...
         return get_object_or_404(WriteItInstance, pk=self.kwargs['pk'])
 
-    def get_object(self):
+    def check_ownership(self):
         self.writeitinstance = self.get_writeitinstance()
         if not self.writeitinstance.owner.__eq__(self.request.user):
             raise Http404
+
+    def get_object(self):
+        self.check_ownership()
```
lfalvarez: naming improvement

PR #496, opened 2015-01-30
Celery testing

lfalvarez:


lfalvarez:
This does not do or fix anything but I got to prove that this new settings work in travis. There are upcoming fixes and features in this branch

Issue #373, opened 2014-12-18
Use cuttlefish


lfalvarez:
Given that cuttlefish has recently [open to be used as a service](http://poplus.org/posts/email-is-your-secret-weapon/) we should move our service to use cuttlefish instead of Amazon ses

tmtmtmtm:
@lfalvarez Do you have an estimate of how much work is still required on this one?

lfalvarez:
I'd say a day or less

paullenz:
Ooo - that is *very* good news!

lfalvarez:
I'm having a slight problem with the encoding in cuttlefish. I've been in contact with Matthew and he'll be seeing it soon.

lfalvarez:
This is pretty much ready and working! =)

tmtmtmtm:
Woohoo!

lfalvarez:
I think it would be worth adding a couple of send the admins an email if everything fails before closing this.

PR #534, opened 2015-02-24
Using custom email config per instance if given

lfalvarez:
related to #373

### DIFF
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -735,9 +735,11 @@ def send_an_email_to_the_author(sender, instance, created, **kwargs):
         if settings.SEND_ALL_EMAILS_FROM_DEFAULT_FROM_EMAIL:
             from_email = settings.DEFAULT_FROM_EMAIL
         else:
+            from_domain = confirmation.message.writeitinstance.config.custom_from_domain\
+                or settings.DEFAULT_FROM_DOMAIN
             from_email = "%s@%s" % (
                 confirmation.message.writeitinstance.slug,
-                settings.DEFAULT_FROM_DOMAIN,
+                from_domain,
                 )
 
         msg = EmailMultiAlternatives(
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -399,11 +399,13 @@ class Message(models.Model):
                 or settings.DEFAULT_FROM_DOMAIN
             from_email = self.writeitinstance.slug + "@" + from_domain
 
+        connection = self.writeitinstance.config.get_mail_connection()
         msg = EmailMultiAlternatives(_('Moderation required for\
          a message in WriteIt'),
             text_content,  # content
             from_email,  # From
-            [self.writeitinstance.owner.email]  # To
+            [self.writeitinstance.owner.email],  # To
+            connection=connection,
             )
         msg.attach_alternative(html_content, "text/html")
         msg.send()
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -151,6 +151,14 @@ class WriteItInstanceConfig(models.Model):
         help_text=_("Messages pushed to the api should \
             be confirmed automatically"), default=True)
 
+    custom_from_domain = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host_password = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host_user = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_port = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_use_tls = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_use_ssl = models.CharField(max_length=512, null=True, blank=True, default=None)
+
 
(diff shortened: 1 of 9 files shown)


nuntium/models.py
```diff
@@ -151,6 +151,25 @@ class WriteItInstanceConfig(models.Model):
...
+    custom_from_domain = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host_password = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_host_user = models.CharField(max_length=512, null=True, blank=True, default=None)
+    email_port = models.IntegerField(null=True, blank=True, default=None)
+    email_use_tls = models.NullBooleanField(default=None)
+    email_use_ssl = models.NullBooleanField(default=None)
+
+    def get_mail_connection(self):
+        from django.core import mail
```
duncanparkes: I'd do this import at the top of the file.

nuntium/tests/confirmation_test.py
```diff
@@ -95,6 +95,38 @@ def test_it_sends_an_email_to_the_author_asking_for_confirmation(self):
...
+        self.assertEquals(connection.port, config.email_port)
+        self.assertEquals(connection.use_tls, config.email_use_tls)
+        '''
+        I'm moving all the site to use cuttlefish but in the meantime
+        in order to test I'm using this specific config per instance
+
+        EMAIL_HOST = 'cuttlefish.oaf.org.au'
+        EMAIL_PORT = 2525
+        EMAIL_HOST_USER = 'writeit'
+        EMAIL_HOST_PASSWORD = 'FieraFerozEsElMejorPerroDelMundo'
```
duncanparkes: This isn't your real password, is it?

lfalvarez:
cool @duncanparkes is it so I can use it somewhere else in the code?

nuntium/tests/confirmation_test.py
lfalvarez: naaa is just an example and

nuntium/tests/confirmation_test.py
```diff
@@ -95,6 +95,38 @@ def test_it_sends_an_email_to_the_author_asking_for_confirmation(self):
         expected_from_email = self.message.writeitinstance.slug + "@" + settings.DEFAULT_FROM_DOMAIN
         self.assertEquals(mail.outbox[0].from_email, expected_from_email)
 
+    def test_sends_confirmation_from_a_custom_domain_if_specified(self):
+        '''Sending confirmation from a custom domain if specified'''
+        config = self.message.writeitinstance.config
+        config.custom_from_domain = "custom.domain.cl"
+        config.email_host = 'cuttlefish.au.org'
+        config.email_host_password = 'f13r4'
+        config.email_host_user = 'fiera'
```
duncanparkes: Ditto this password I guess, though this is short and looks fake :-)

nuntium/tests/confirmation_test.py
duncanparkes: Phew! It just looked long and complicated enough that it could have been a real one in there by accident :-)

nuntium/models.py
```diff
@@ -151,6 +152,24 @@ class WriteItInstanceConfig(models.Model):
         help_text=_("Messages pushed to the api should \
             be confirmed automatically"), default=True)
 
+    custom_from_domain = models.CharField(max_length=512, null=True, blank=True, default=None)
```
duncanparkes: I wonder if the `default=None`s are really necessary here? Is that not the default default?

nuntium/models.py
```diff
@@ -727,16 +746,21 @@ def send_an_email_to_the_author(sender, instance, created, **kwargs):
         if settings.SEND_ALL_EMAILS_FROM_DEFAULT_FROM_EMAIL:
             from_email = settings.DEFAULT_FROM_EMAIL
         else:
+            from_domain = settings.DEFAULT_FROM_DOMAIN
```
duncanparkes: You could write this as an or, using the short-circuit:

`from_domain = confirmation.message.writeitinstance.config.custom_from_domain or settings.DEFAULT_FROM_DOMAIN`

nuntium/models.py
```diff
@@ -727,16 +746,21 @@ def send_an_email_to_the_author(sender, instance, created, **kwargs):
...
+                from_domain,
                 )
+        connection = confirmation.message.writeitinstance.config.get_mail_connection()
 
         msg = EmailMultiAlternatives(
             subject,
             text_content,  # content
             from_email,  # From
             [confirmation.message.author_email],  # To
+            connection=connection
```
duncanparkes: I like final commas on lines like this so that when I come back to it and add another parameter I don't end up giving myself a syntax error!

nuntium/models.py
lfalvarez: We could give it a go! But I was just a bit afraid that south would say something like "What do you want to do with backwards migrations?", so in order to avoid that sort of thing I just declare it. But it doesn't mean that it isn't the default by default.

nuntium/models.py
lfalvarez: Cool!

PR #551, opened 2015-03-02
Add list of recipients to message admin page.

### LINKED ISSUE #288: Could the recipient be added as a field on the message page in the admin interface
It would be good to list the recipient on the as a field on the message page  - http://writeit.ciudadanointeligente.org/admin/nuntium/message/1055/

duncanparkes:
Fixes #288


lfalvarez:
looks simple enough to me as well

PR #552, opened 2015-03-02
No message creation form if no contacts.

### LINKED ISSUE #401: Don't provide contact form if an instance has no contacts
If an instance has no contacts set up, we shouldn't provide a form that lets you write to people (but with a zero size list of possible contacts!)

e.g. http://writeit.ciudadanointeligente.org/en/writeit_instances/legislativo-chile

duncanparkes:
If there are no available contacts, say that there is no-one to
write to rather than displaying a form for creating a message.
Obviously this all needs design tidying up.

Fixes #401.


lfalvarez:
looks pretty good to me :+1:

PR #553, opened 2015-03-02
Add post submission page 198

### LINKED ISSUE #198: Add page post submission
Currently when the user has submitted they are taken back to the home page with just a message at the top.  Could we instead have a separate page with a strong call to action which gives the 'Check you email message' and smaller links to 'send another message' and 'view messages and replies'

lfalvarez:
this closes #198

### DIFF
--- a/nuntium/templates/frontend/post_submission.html
+++ b/nuntium/templates/frontend/post_submission.html
@@ -1,0 +1,14 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<div class="jumbotron">
+    <div class="container">
+        <h1>{% trans "Your message has been created!" %}</h1>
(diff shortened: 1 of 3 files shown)


duncanparkes:
Hi @lfalvarez 

I'm a bit confused about what you're doing here. Are you replacing little message at the top of the screen with a rendered template containing a fairly similar message which then automatically redirects after a while to the original place? I'm not sure what the advantage of that is.

I was expecting a new view and url which would be used as the success_url on creation of a message. It looks like Paul wanted that page to have several links on it. It would be odd for it to then automatically redirect to somewhere else while someone was considering clicking on them!

Cheers,

Duncan

lfalvarez:
hey @duncanparkes ! I think I agree with you I'm going to take another approach =)

Issue #453, opened 2015-01-15
META: 'Message Composer' page


tmtmtmtm:
- [x] Should be on a page of its own
- [x] Don't show the "Trailing * means that we don't have how to contact the person" text [#677]
- [x] Don't display the form at all if there are no contacts [#401] 
- [x] Don't display the form at all if set to only accept messages via the API [#716]
- [x] Remove the public/private toggle [#436]
- [ ] Explain what the process is going to be (e.g. that they'll need to confirm by email)

tmtmtmtm:
I'm happy with this page!

PR #556, opened 2015-03-03
#198 add page post submission

lfalvarez:

### DIFF
--- a/nuntium/templates/nuntium/message/post_submission.html
+++ b/nuntium/templates/nuntium/message/post_submission.html
@@ -4,7 +4,7 @@
 {% block content %}
 <div class="jumbotron">
     <div class="container">
-        <h1>{% trans "Your message has been created!" %}</h1>
+        <h1>{% trans "Please check your email for a confirmation link" %}</h1>
         {% if messages %}
             {% for message in messages %}
               <p>{{ message }}</p>
--- a/nuntium/templates/nuntium/message/post_submission.html
+++ b/nuntium/templates/nuntium/message/post_submission.html
@@ -5,10 +5,8 @@
 <div class="jumbotron">
     <div class="container">
         <h1>{% trans "Please check your email for a confirmation link" %}</h1>
-        {% if messages %}
-            {% for message in messages %}
-              <p>{{ message }}</p>
-            {% endfor %}
+        {% if moderation_follows %}
+          <p>{% trans "After you confirm your message will be waiting form moderation" %}</p>
         {% endif %}
         <p><a href="{% url 'instance_detail' slug=message.writeitinstance.slug %}" class="btn btn-primary btn-lg">{% blocktrans %}Create another message and view messages and replies{% endblocktrans %}<i class="icon-double-angle-right"></i></a></p>
     </div>
--- a/nuntium/templates/nuntium/message/post_submission.html
+++ b/nuntium/templates/nuntium/message/post_submission.html
@@ -1,0 +1,16 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<div class="jumbotron">
+    <div class="container">
+        <h1>{% trans "Your message has been created!" %}</h1>
+        {% if messages %}
+            {% for message in messages %}
+              <p>{{ message }}</p>
+            {% endfor %}
+        {% endif %}
+        <p><a href="{% url 'instance_detail' slug=message.writeitinstance.slug %}" class="btn btn-primary btn-lg">{% blocktrans %}Create another message and view messages and replies{% endblocktrans %}<i class="icon-double-angle-right"></i></a></p>
+    </div>
+</div>
+{% endblock content %}
--- a/nuntium/tests/writeitinstances_test.py
+++ b/nuntium/tests/writeitinstances_test.py
@@ -316,14 +316,9 @@ class InstanceDetailView(TestCase):
 
         response = self.client.post(self.url, data, follow=True)
         self.assertEquals(response.status_code, 200)
-        expected_acknowledgement = _("Thanks for submitting your message, please check your email and click on the confirmation link")
 
-        self.assertContains(response, expected_acknowledgement)
-
-        all_messages, all_retrieved = response.context["messages"]._get()
-
-        self.assertEquals(len(all_messages), 1)
-        self.assertEquals(all_messages[0].__str__(), expected_acknowledgement)
+        self.assertIn('moderation_follows', response.context)
+        self.assertFalse(response.context['moderation_follows'])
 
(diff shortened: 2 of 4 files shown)


lfalvarez:
This is related to #198

nuntium/templates/nuntium/message/post_submission.html
```diff
@@ -0,0 +1,16 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<div class="jumbotron">
+    <div class="container">
+        <h1>{% trans "Your message has been created!" %}</h1>
+        {% if messages %}
```
duncanparkes: Are you getting two things mixed up here? This template variable 'messages' is from the Django Messaging Framework (https://docs.djangoproject.com/en/1.7/ref/contrib/messages/), isn't it? So it's just the general purpose code for displaying flash messages across the top of whatever page is being viewed - normally I'd expect this stuff to be in a base template so that it appears everywhere.

Or have I got the wrong end of the stick?

nuntium/tests/writeitinstances_test.py
```diff
@@ -302,9 +302,8 @@ def test_message_creation_on_post_form(self):
         self.assertEquals(response.status_code, 200)
         new_messages = Message.objects.filter(subject='Fiera no está')
         self.assertTrue(new_messages.count() > 0)
-        self.assertEquals(len(response.context["form"].errors), 0)
 
-    def test_I_get_an_acknoledgement_for_creating_a_message(self):
+    def test_get_an_acknoledgement_for_creating_a_message(self):
```
duncanparkes: Might as well correct the spelling in the method name to 'acknowledgement' while you're changing it.

nuntium/tests/writeitinstances_test.py
```diff
@@ -372,3 +378,25 @@ def test_no_form_on_homepage_of_empty_instance(self):
...
         self.assertIn('there is no-one to write to', response.content)
+
+    def test_there_is_a_post_submission_url(self):
+        '''There is a post submission URL'''
+        message = self.writeitinstance1.message_set.get(id=1)
+        url = reverse('post_submission', kwargs={
+            'instance_slug': self.writeitinstance1.slug,
+            'slug': message.slug,
+            })
+        self.assertTrue(url)
```
duncanparkes: I don't think there's any point in having this assertion here - if reverse can't find a URL, it will raise ` NoReverseMatch`, so this won't be executed.

nuntium/tests/writeitinstances_test.py
```diff
@@ -372,3 +378,25 @@ def test_no_form_on_homepage_of_empty_instance(self):
...
+        message = self.writeitinstance1.message_set.get(id=1)
+        url = reverse('post_submission', kwargs={
+            'instance_slug': self.writeitinstance1.slug,
+            'slug': message.slug,
+            })
+        response = self.client.get(url)
+        self.assertEquals(response.status_code, 200)
+        self.assertIn('message', response.context)
+        self.assertEquals(response.context['message'], message)
+        self.assertTemplateUsed(response, 'nuntium/message/post_submission.html')
```
duncanparkes: I think it would be nice to include a test which uses the test client to post to create a new message and checks for the redirect.

nuntium/views.py
```diff
@@ -92,14 +96,21 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class MessageDetailView(DetailView):
-    template_name = 'nuntium/message/message_detail.html'
+class MessageDetailViewMixin(DetailView):
```
duncanparkes: I'd call this `MessageDetailBaseView` - it's not really being used as a mixin.

nuntium/templates/nuntium/message/post_submission.html
lfalvarez: Yep you're right, that's true. I think it could be a good idea to move this logic https://github.com/ciudadanointeligente/write-it/blob/198-add-page-post-submission/nuntium/views.py#L72-L75 over to the template rather than using the Django Messaging Framework

nuntium/tests/writeitinstances_test.py
lfalvarez: Cool! I'll do so!

nuntium/tests/writeitinstances_test.py
lfalvarez: Agree, but it made some sense when I was developing, I think it is jut something that I'm used to do when I'm coding.

nuntium/tests/writeitinstances_test.py
lfalvarez: I think this test does so https://github.com/ciudadanointeligente/write-it/blob/198-add-page-post-submission/nuntium/tests/writeitinstances_test.py#L328-L346

nuntium/views.py
```diff
@@ -92,14 +96,21 @@ def get_context_data(self, **kwargs):
...
     model = Message
 
     def get_queryset(self):
```
duncanparkes: I think you could override `get_object` here rather than `get_queryset`. What you really want is to find the unique Message object which has slug `self.kwargs['slug']`, isn't it? I'm currently confused as to how PostSubmissionView finds its object.

lfalvarez: Uhh yeah, you're right! I hadn't seen that. Well there's something else related to the url cause what it does is include the instance_slug https://github.com/ciudadanointeligente/write-it/blob/198-add-page-post-submission/nuntium/subdomain_urls.py#L17, and that's why it was originally intended to use this get_queryset. But I could definately move this over to the get_object method.

duncanparkes: That's a good idea - the `get_object` method could include writeitinstance__slug=kwargs['instance_slug'] in it's filters just to make sure that the message is related to the instance that it is being requested from.

nuntium/tests/writeitinstances_test.py
duncanparkes: Oh so it does - I somehowe missed that when reading the diff. Sorry!

lfalvarez: No worries =)

nuntium/templates/nuntium/message/post_submission.html
duncanparkes: I agree. The fact that the message creator needs to check their email and click the confirmation link is the reason why I think we need a separate page - I was worried that that info might get lost and not acted upon as just a message on the instance homepage. It really needs to be in the main body of the page that you get to after sending a message.

lfalvarez:
I'm wondering how this should be related to https://github.com/ciudadanointeligente/write-it/issues/453, should we wait?

lfalvarez:
This becomes obsolete after the branch ```new_designs```

PR #560, opened 2015-03-04
Issues/444 merge your popit with relate with popit

duncanparkes:
This is all about removing stuff.

1) Get rid of the your PopIt API instances page - it's now covered by the PopIt API stuff on an instance.
2) Get rid of the your contacts page - this is also now dealt with by the instance.
3) Tidy up and remove some text on the form for relating a PopIt to a WriteIt.


lfalvarez:
It looks pretty good to me :+1:

PR #565, opened 2015-03-05
issue/557 handle attachments

lfalvarez:
When an email comes in with an attachment it processes and makes it available for download in the Message detail page.

This is how it is displayed.

Imgur


tmtmtmtm:
@lfalvarez Any chance you can fill in a narrative summary of what this does (or doesn't) do in the description? This is a pretty big change, and I'd find very useful to know what's here (and it'll take me ages to work it out from the code!)

duncanparkes:
I made a couple of small changes before I merged this, which I forgot to comment here first

1) Instead of putting all attachments in a single directory instead I've put them in subdirectories based on date - this is to avoid a future problem of a directory with too many things in it for the filesystem.

2) On one of the tests I replaces an OutboundMessageIdentifier.objects.get(id=1) with a fetch of the outboundmessageidentifier attribute of an OutboundMessage object we already had. I was seeing test failures because OutboundMessage 1 didn't always have the OutboundMessageIdentifier number 1 associated with it.

lfalvarez:
:+1:

PR #573, opened 2015-03-09
428 handle more than text plain responses

lfalvarez:
This PR parses HTML responses, stores it, but it does not display it, just yet. The problem here is that the HTML part of an email will contain the signature part, opposed to the plain text for which I'm using the [email-reply-parser](https://github.com/zapier/email-reply-parser) lib.
This PR also contains the structure necesary for parsing 'multipart/alternative' parts, this part normally acts as a container of the other 2 important ones: text/plain and text/html.

### DIFF
--- a/mailit/bin/handleemail.py
+++ b/mailit/bin/handleemail.py
@@ -109,6 +109,9 @@ class EmailHandler():
     def __init__(self, answer_class=EmailAnswer):
         self.message = None
         self.answer_class = answer_class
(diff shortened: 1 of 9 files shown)


lfalvarez:
Please do not merge yet, until I figure out the way for managing 'multipart/alternative'.

lfalvarez:
I'm closing this in favour of #575 which is a simpler approach.

PR #575, opened 2015-03-10
Handle text html

### LINKED ISSUE #428: Handle more than text/plain replies
We currently don't handle:
- [ ] text/html
- [ ] multipart/mixed
- [ ] multipart/alternative

lfalvarez:
This together with #565 should close #428.
This PR adds the ability to store html content of an email. Currently it is not being displayed as the html content would include the signature of the sender, which is really hard to get rid of (for now).

### DIFF
--- a/mailit/bin/handleemail.py
+++ b/mailit/bin/handleemail.py
@@ -67,6 +67,7 @@ class EmailAnswer(EmailSaveMixin, EmailReportBounceMixin):
     def __init__(self):
         self.subject = ''
         self._content_text = ''
+        self.content_html = ''
         self.outbound_message_identifier = ''
         self.email_from = ''
         self.when = ''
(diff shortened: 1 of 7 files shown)


lfalvarez:
Are we using any of this code?

duncanparkes:
I'm not sure what you mean. I hope we're using it, yes!

I was meaning that it should be OK to just rename the migrations as they're all pretty independent, if that's what you're referring to?

duncanparkes:
Ah - do you mean because it has unmerged commits? I rebased before I merged, so there are new commits which look the same - it's all on master now.

lfalvarez:
oh ! really!?? Cool!

lfalvarez:
I got kind of lost somewhere =)

duncanparkes:
I squashed it all together and added a couple of extra bits, but it's all there at the top of master as commit 228cee09f022fd1b692e805986dc0cd6a46bc5b4

Let me know if I've somehow lost something, of course!

lfalvarez:
no worries =)

PR #479, opened 2015-01-21
Try using env to test on multiple backends in travis

duncanparkes:
Don't merge yet, I'm just trying to test something on travis.

PR #576, opened 2015-03-10
Picking objects by id rather than getting all of them and choosing the first one from the queryset.

lfalvarez:
What I did was changing several tests to use the method ```.get()``` rather than getting all the objects ```.all()``` and choosing the first of the queryset.

This PR should fix some problems regarding tests failing in an intermitent way. The problem was described by @duncanparkes in #479. (Sorry it took me this long to get to this point)

This is an specially big change to the tests, particularly 774bcb77bdbd9dead20a13aa2cb511ff1c57ff83, and of course can be split in smaller changes.


mailit/tests/plugin_test.py
```diff
@@ -109,11 +109,11 @@ def test_it_only_creates_templates_when_creating_not_when_updating(self):
 class MailSendingTestCase(TestCase):
     def setUp(self):
         super(MailSendingTestCase, self).setUp()
-        self.person3 = Person.objects.all()[2]
+        self.person3 = Person.objects.get(id=1)
```
duncanparkes: Is this getting the same object? And does it matter?

nuntium/tests/answers_test.py
```diff
@@ -7,10 +7,10 @@
 class AnswerTestCase(TestCase):
     def setUp(self):
         super(AnswerTestCase, self).setUp()
-        self.message = Message.objects.all()[0]
-        self.person = Person.objects.all()[0]
+        self.message = Message.objects.get(id=1)
+        self.person = Person.objects.get(id=1)
         #There is no membership for this guy to any writeitInstance
-        self.person_not_in_the_instance = Person.objects.all()[1]
+        self.person_not_in_the_instance = Person.objects.get(id=2)
```
duncanparkes: And this one? I'm just slightly concerned because I was assuming you were getting them back in primary key order in your environment.

duncanparkes:
Thanks for chasing all these down.

mailit/tests/plugin_test.py
duncanparkes: I'm not sure why the variable is called person3, rather than person1, but everything passes for me repeatedly, which is nice, so I'm going to merge.

lfalvarez: I've just checked and it doesn't matter, I think it is worth changing the name of the variable to just ```self.person``` or something simpler.

PR #588, opened 2015-03-13
Removing the link for seing the public non-confirmed messages and the co...

### LINKED ISSUE #528: Unconfirmed messages 404 from the admin page
From the admin interface, there are links to the 'internal' and 'external' views of a message. However, if the message hasn't been confirmed yet, it currently 404s.

Either: we should make it so that an admin can see unconfirmed messages at the public URL (with something prominent noting that it hasn't been confirmed yet [and probably a link to confirm it])

Or: we should disable that link on the admin page.

lfalvarez:
...lumn with the public link. This should close #528.


lfalvarez:
I'm very open to suggestions!!!

This is how it looks like:

PR #590, opened 2015-03-16
Add a Vagrantfile for running WriteIt in a Vagrant VM

zarino:
This includes:
* Simple Vagrantfile with NFS shared directory
* Provisioning script with zero-input django migrations
* Removal of PostgreSQL-specific package (psycopg2) in requirements.txt
* Documentation in README.md
* Steps to run webserver also printed out on SSH login

Vagrant file tested on a Mac. Can't see any reason why it wouldn't also work on Linux.

@duncanparkes – is there any other setup that the Vagrant provisioner could do, to get people up and running quicker?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,16 +5,38 @@ You write it, and we deliver it.
 [![Coverage Status](https://coveralls.io/repos/ciudadanointeligente/write-it/badge.png?branch=master)](https://coveralls.io/r/ciudadanointeligente/write-it)
 [![Code Health](https://landscape.io/github/ciudadanointeligente/write-it/master/landscape.png)](https://landscape.io/github/ciudadanointeligente/write-it/master)
 
-Write-it is an application that aims to deliver messages to people whose contacts are to be private or the messages should be public, for example: members of congress. 
+Write-it is an application that aims to deliver messages to people whose contacts are to be private or the messages should be public, for example: members of congress.
 
 Write-it is a layer on top of [popit](http://popit.mysociety.org) from where it takes the people and adds contacts. The way it delivers messages is using plugins for example: mailit. And this approach allows for future ways of delivering for example: twitter, whatsapp, fax or pager.
 
 Future uses are in [congresoabierto](http://www.congresoabierto.cl) to replace the old "preguntales" (You can [check here](http://congresoabierto.cl/comunicaciones), to see how it used to work) feature, it was inspired by [writetothem](http://www.writetothem.com/).
 
 
+Quick Installation (Vagrant)
+============================
 
-Installation
-------------
+Assuming [you have Vagrant installed](http://docs.vagrantup.com/v2/installation/), run the following:
+
+    git clone https://github.com/ciudadanointeligente/write-it.git
+    cd write-it
+    vagrant up
+
+Vagrant will automatically install WriteIt and all of its dependencies. This can take a few minutes.
+
+Once it’s complete, log into the virtual machine with:
+
+    vagrant ssh
+
+Once you’re inside the virtual machine, run the web server with:
+
+    cd /vagrant
+    ./manage.py runserver 0.0.0.0:8000
+
+And visit http://localhost:8000 on your host machine to use WriteIt.
+
+
+Manual Installation (without Vagrant)
+=====================================
 
 System Requirements
 -------------------
@@ -32,8 +54,6 @@ System Requirements
  * Libssl
 
  In ubuntu you can do ```sudo apt-get install libssl-dev```
-
-
 
 Write-it is built using Django. You should install Django and its dependencies inside a virtualenv. We suggest you use [virtualenvwrapper](http://virtualenvwrapper.readthedocs.org/en/latest/) to create and manage virtualenvs, so if you don’t already have it, [go install it](http://virtualenvwrapper.readthedocs.org/en/latest/install.html#basic-installation), remembering in particular to add the required lines to your shell startup file.
 
@@ -53,7 +73,7 @@ Set up the database, creating an admin user when prompted:
 
 Troubleshooting database migration
 ----------------------------------
-There's a problem migrating and the problem looks like 
+There's a problem migrating and the problem looks like
 
 	django.db.utils.OperationalError: no such table: tastypie_apikey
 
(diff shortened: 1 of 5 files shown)


scripts/provision.sh
```diff
@@ -0,0 +1,36 @@
...
+cd /vagrant
+sudo pip install --timeout=120 --use-mirrors -r requirements.txt
+
+# Set up the Django database
+./manage.py syncdb --noinput
+./manage.py migrate
+
+# Set on-login.sh to be run when the user logs in
+echo '' >> ~/.profile
+echo '/vagrant/scripts/on-login.sh' >> ~/.profile
```
duncanparkes: There's no need to blank the .profile separately if you do 

`echo '/vagrant/scripts/on-login.sh' > ~/.profile`

with a single `>`.

duncanparkes:
There is a script at the root of the git checkout which starts a little local Popit API thing for testing.

start_local_popit_api.bash

This runs on port 3000, so I guess you might want to start that up and forward the port (though perhaps you don't need to forward the port for running the unit tests as it'll all be inside the vagrant box).

scripts/provision.sh
zarino: @duncanparkes – Using a single `>` will remove everything that's already in the `.profile` – are you sure you want to do that?

The reason for my line 31 (`echo '' >> ~/.profile`) is just to add a blank line between the existing content of `.profile` and my new line at the end.

README.md
```diff
@@ -5,16 +5,38 @@ You write it, and we deliver it.
...
 
-Installation
-------------
+Assuming [you have Vagrant installed](http://docs.vagrantup.com/v2/installation/), run the following:
+
+    git clone git@github.com:ciudadanointeligente/write-it.git
+    cd write-it
+    vagrant up
+
+Vagrant will automatically install WriteIt and all of its dependancies. This can take a few minutes.
```
chrismytton: Typo here "dependancies" should be "dependencies"

README.md
```diff
@@ -5,16 +5,38 @@ You write it, and we deliver it.
...
 
 
+Quick Installation (Vagrant)
+============================
 
-Installation
-------------
+Assuming [you have Vagrant installed](http://docs.vagrantup.com/v2/installation/), run the following:
+
+    git clone git@github.com:ciudadanointeligente/write-it.git
```
chrismytton: Probably best to use the http url here since it's a public repo, save people having to worry about ssh keys etc

```
git clone https://github.com/ciudadanointeligente/write-it.git
```

scripts/provision.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo
+echo '** Your Vagrant box is ready to use! \o/ **'
+echo 'Log in (with vagrant ssh) and follow the instructions.'
```
chrismytton: Can we put these lines into a the [`config.vm.post_up_message`](http://docs.vagrantup.com/v2/vagrantfile/machine_settings.html) so they're shown every time the VM is started? The same as [popit does](https://github.com/mysociety/popit/blob/3f715bd56e92650dee8888218d8d3b33ad845655/Vagrantfile#L34).

scripts/provision.sh
chrismytton: The alternative (and arguably more correct way!) would be to add the text from `on-login.sh` directly into `/etc/motd.tail` so it gets shown on login? By default `/etc/motd.tail` contains "Welcome to your Vagrant-built virtual machine.", so it seems appropriate to put some extra info in there.

Vagrantfile
```diff
@@ -0,0 +1,25 @@
+# Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
+VAGRANTFILE_API_VERSION = "2"
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+  # Every Vagrant virtual environment requires a box to build off of.
+  config.vm.box = "precise64"
```
chrismytton: If you change this to

```
config.vm.box = "hashicorp/precise64"
```

then you can drop the `box_url` below and it will just download it from vagrant cloud if it can't find the box locally. I think this is the recommended way of doing things these days.

Vagrantfile
```diff
@@ -0,0 +1,25 @@
...
+
+  # The url from where the 'config.vm.box' box will be fetched if it
+  # doesn't already exist on the user's system.
+  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
+
+  # Enable NFS access to the disk
+  config.vm.synced_folder "", "/vagrant", :nfs => true
```
chrismytton: The rubyist in me wants this to be `nfs: true` to match the style in the rest of the file, but obviously not essential!

chrismytton:
This is great, thanks @zarino! I've added a few general comments, but the main feature (running `vagrant up`) does work correctly, which is the important thing!

lfalvarez:
:+1: thats awesome!

Vagrantfile
zarino: Ah, I think I copied a Vagrantfile written by a mySociety Services pythonist ;-)

scripts/provision.sh
zarino: This is exactly what I wanted!! I spent ages looking up a way to print something when the machine gets upped! Eventually settled for this. `config.vm.post_up_message` is way better!

scripts/provision.sh
zarino: :trophy: to @chrismytton for his awesome unix knowledge. I'll see what it's `/etc/motd.tail` by default, and either append to it, or overwrite it, in the provisioning script.

zarino:
@duncanparkes – I've addressed @chrismytton's comments on the code. I think this is ready to merge.

Tomorrow, we should try running the tests inside the VM, and see whether your port 3000 redirect is required.

zarino:
(PS. feel free to squash https://github.com/ciudadanointeligente/write-it/commit/478e50eaf4d613d466cd293dd6ab36000c4eeaec into https://github.com/ciudadanointeligente/write-it/commit/1b4e3a28da30d31b0d92d44fe042d53d9a4a7288)

Issue #200, opened 2013-08-22
"Report this message" button


paullenz:
Could we add a button against both public message and responses for users to report message that they believe to be offensive etc

The admin interface will list all reported message with the option available for the admin user to hide or mark as okay as appropriate

lfalvarez:
Here I think it would be interesting to understand something:
What if a user wanted to hide a message for no reason. Could he do so? if a user could then the action of hiding a message should be in a separate issue.

tmtmtmtm:
Anyone should be able to *report* a message as needing attention, but only the administrator will be able to hide them.

chrismytton:
Could this potentially be done with a simple `mailto:` link which just emails admin@writeitpublic.com or whatever?

tmtmtmtm:
@chrismytton :+1:

lfalvarez:
I think we could maybe close this in favour of #713

PR #603, opened 2015-03-18
Report this message button

lfalvarez:
This PR does:
- Adds a model called Report related to a message with the "reason" and "datetime" attributes, so the owner of the instance knows a bit more about why it was reported.
- Adds a view and a url for creating reports per messages.
- Displays the existance of reports and details about the report in the backend.

This PR does not:
- Allow a user to hide a message if she/he thinks is necessary.
- Display the link for adding a report, as it is hidden.

Questions regarding #200: 
- A user can hide a message for no reason?
- How should we display a report form?


lfalvarez:
I'm closing this as probably it is an overkill compared to @chrismytton's suggestion of using a simple ```mailto:```

PR #612, opened 2015-03-19
See all messages from a person

### LINKED ISSUE #442: See all messages from a person
Even though there's no actual user account, we can simulate this, as they're all from confirmed email addresses.

lfalvarez:
This PR does:
* Provide a view to see all messages sent by someone.
It does so by providing a link of the form  ```<slug>/from/<message_slug>``` and using the message author_email to search between other messages in the same instance.
This should close #442


lfalvarez:
Probably using a hash as discussed with @duncanparkes and @tmtmtmtm would produce canonical urls and is a better approach for this.

nuntium/templates/nuntium/message/message_detail.html
```diff
@@ -17,7 +17,9 @@ <h1 class="panel-title">{{ message.subject }}</h1>
     </div>
   </div>
   <div class="panel-footer row">
-    <span class="author_name col-md-4"><small>{% blocktrans with message.author_name as name %}Asked by {{ name }}.{% endblocktrans %}</small></span>
+    {% url 'all-messages-from-the-same-author-as' slug=message.writeitinstance.slug message_slug=message.slug as other_messages_by %}
+    {% blocktrans with message.author_name as name %}Asked by 
+    <a href="{{other_messages_by}}">{{ name }}</a>.{% endblocktrans %}</small></span>
```
duncanparkes: I think you'll need to put `other_messages_by` into the `with` in the `blocktrans` too.

nuntium/templates/nuntium/message/message_in_list.html
```diff
@@ -28,7 +28,10 @@
 	{% endfor %}
 
 	<br />
-	<span class="author_name col-md-4"><small>{% blocktrans with message.author_name as name %}Asked by {{ name }}.{% endblocktrans %}</small></span>
+	<span class="author_name col-md-4"><small>
+    {% url 'all-messages-from-the-same-author-as' slug=message.writeitinstance.slug message_slug=message.slug as other_messages_by %}
+    {% blocktrans with message.author_name as name %}Asked by 
+    <a href="{{other_messages_by}}">{{ name }}</a>.{% endblocktrans %}</small></span>
```
duncanparkes: Ditto here - `other_messages_by` needs to be in the with of the `blocktrans` unless I've misunderstood how blocktrans works (which I could well have).

nuntium/templates/nuntium/message/from_person.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+
+{% block content %}
+<h1>{% blocktrans with person_name=author_name%}Messages sent by {{person_name}} {% endblocktrans %}</h1>
```
duncanparkes: Why not stick with just `author_name` in the blocktrans too?

nuntium/views.py
```diff
@@ -214,3 +215,25 @@ def get_context_data(self, **kwargs):
...
+
+    def dispatch(self, *args, **kwargs):
+        self.writeitinstance = get_object_or_404(WriteItInstance, slug=self.kwargs['slug'])
+        self.message = get_object_or_404(Message, slug=self.kwargs['message_slug'])
+        return super(MessagesFromPersonView, self).dispatch(*args, **kwargs)
+
+    def get_queryset(self):
+        qs = super(MessagesFromPersonView, self).get_queryset()
+        qs = qs.filter(writeitinstance=self.writeitinstance).\
+            filter(public=True).filter(confirmated=True).\
```
duncanparkes: Do we now have a single way of checking to see if something is confirmed, or are there still two different ways? I seem to remember there was both a `confirmated` column and the existence of a confirmation object to check.

duncanparkes:
Is writeit_test_1.sqlite3 included deliberately?

duncanparkes:
I don't think we should worry about the canonical url thing right now - we can always change the urls this PR starts with to redirects to a canonical one if we make one without doing any harm.

If this is merged after #618, then we'll have to be careful about the urls.

duncanparkes:
I think if we sort out:

 [ ] The blocktrans things
 [ ] Check the confirmation situation
 [ ] Check the binary file should actually be there

Then this is ready to go.

nuntium/templates/nuntium/message/message_in_list.html
lfalvarez: I thought so too but apparently the way you define a url to be used in a blocktrans is by previously call the ```{% url as name %}``` filter and then call the blocktrans, according to this [stackoverflow](http://stackoverflow.com/questions/11271002/django-trans-and-url-tags).

nuntium/views.py
lfalvarez: For some reason a long time ago I created the confirmated field which should be transformed into a ```@property```. But I also think it should go into a separate issue. What do you say?

lfalvarez:
I'm wondering what is the best url for this porpuse? I think that there is a conflict with master regarding specifically that.

nuntium/templates/nuntium/message/message_in_list.html
duncanparkes: OK - so I have misunderstood blocktrans :-)

duncanparkes:
:+1: Looks good!

PR #617, opened 2015-03-20
Issues/609 configurable number of recipients

lfalvarez:
This PR adds a new field to the WriteitInstanceConfig model named maximum_recipients and this field regulates how many recipients you send a message at once.

By default this number is 5 and can be moved up to 10 by the user in the admin section.

This number can be set higher if you speak to the person in charge of the system.


lfalvarez:
uhh before mergin this contains a migration which probably needs to be rebuild

duncanparkes:
:+1: 

Just one thing - you'll need to re-number the migration now you've merged #628. It should still work though as the migration there was only a data migration.

PR #619, opened 2015-03-23
Extracting from popit adds basic logic to enable or disable contacts.

### LINKED ISSUE #419: Set default 'active' flag per recipient depending on Membership status
a Person who only has Memberships that are no longer active, should be disabled by default.

lfalvarez:
If any Membership is still active, then they should be enabled. This should probably close #419.

### DIFF
--- a/nuntium/popit_api_instance.py
+++ b/nuntium/popit_api_instance.py
@@ -1,6 +1,27 @@
 from popit.models import ApiInstance, Person, get_paginated_generator
 from contactos.models import Contact
 from mailit import MailChannel
+from datetime import datetime
+
+
+def get_date_or_none(membership_doc, key, format="%Y-%m-%d"):
+    try:
+        return datetime.strptime(membership_doc.get(key, None), format)
+    except TypeError:
+        return None
+
+
+def _is_current_membership(start_date, end_date):
+    today = datetime.today()
+    is_valid = ((start_date is None) or start_date <= today) and\
+        ((end_date is None) or end_date >= today)
+    return is_valid
+
+
+def is_current_membership(membership_doc, start_date_key='start_date', end_date_key='end_date'):
+    start_date = get_date_or_none(membership_doc, start_date_key)
+    end_date = get_date_or_none(membership_doc, end_date_key)
+    return _is_current_membership(start_date, end_date)
 
 
 class PopitPerson(Person):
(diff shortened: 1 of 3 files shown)


nuntium/popit_api_instance.py
```diff
@@ -29,9 +54,15 @@ def fetch_all_from_api(cls, instance, writeitinstance):
             PopitPerson.create_contact(obj, doc, writeitinstance)
 
     @classmethod
+    def determine_if_person_is_current(cls, doc):
+        enable = any([cls.is_active(membership) for membership in doc['memberships']])
```
duncanparkes: I'd just return this rather than having a local variable.

nuntium/popit_api_instance.py
```diff
@@ -1,9 +1,34 @@
...
+
+    @classmethod
+    def _is_active(cls, start_date, end_date):
+        today = datetime.today()
+        is_valid = ((start_date is None) or start_date <= today) and\
+            ((end_date is None) or end_date >= today)
+        return is_valid
+
+    @classmethod
+    def is_active(cls, doc, start_date_key='start_date', end_date_key='end_date'):
```
duncanparkes: It took me a while to realise that `doc` was going to be a membership, from a json file. Perhaps give it a different name to distinguish it from the `doc` in the method below (how about `membership`?)

nuntium/popit_api_instance.py
```diff
@@ -1,9 +1,34 @@
 from popit.models import ApiInstance, Person, get_paginated_generator
 from contactos.models import Contact
 from mailit import MailChannel
+from datetime import datetime
 
 
-class PopitPerson(Person):
+class IsActiveMixin(object):
+    format = "%Y-%m-%d"
```
duncanparkes: This feels like a slightly dangerous attribute name to add to the mixin. You can imagine wanting to use this mixin on something else that already had a `format` attribute to use for something else.

nuntium/popit_api_instance.py
```diff
@@ -29,9 +54,15 @@ def fetch_all_from_api(cls, instance, writeitinstance):
             PopitPerson.create_contact(obj, doc, writeitinstance)
 
     @classmethod
+    def determine_if_person_is_current(cls, doc):
```
duncanparkes: This could actually be a function rather than a class method too. I think putting all this on the class perhaps just complicates things - you can imagine a class you wanted to mix this in to already having an is_active() method on it. I would just have a couple of functions:

```
def is_current_membership(membership_doc):
    ...

def is_current_person(person_doc):
    return any(is_current_membership(membership) for membership in person_doc.get('memberships', ()))
```

nuntium/popit_api_instance.py
lfalvarez: :+1:

duncanparkes:
:+1: I'm happy with this if you merge in the fixup.

Issue #623, opened 2015-03-24
You can push as many messages as you want through the API


lfalvarez:
And a bot can create several messages.

lfalvarez:
Approaches which we could use to fix this problem:

* [ ] [throttling](http://django-tastypie.readthedocs.org/en/latest/throttling.html).
* [ ] Disable API by admin/superadmin.

martinszy:
API keys for individual messages?
You have to obtain the key before displaying the form. This will double api
requests and will get us into a semi-authenticated scheme I'm not very fond
of.

lfalvarez:
do you have somewhere where to read? I wouldn't know how to do it

martinszy:
I just searched for ror api keys and these articules turned up:
http://railscasts.com/episodes/352-securing-an-api?view=asciicast
http://blog.joshsoftware.com/2014/05/08/implementing-rails-apis-like-a-professional/

They are on topic, but I haven't read in detail so I can't assure every
recommendation is worth implementing. We can discuss further or maybe just
implement the other approaches you were proposing.

Martín.

2015-03-24 18:44 GMT+00:00 Luis Felipe Álvarez Burgos <
notifications@github.com>:

> do you have somewhere where to read? I wouldn't know how to do it
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ciudadanointeligente/write-it/issues/623#issuecomment-85638203>
> .
>



-- 
*Martín Szyszlican*
Desarrollo web usable y accesible
martinszyszlican.com

PR #624, opened 2015-03-24
Bug/623 bots can push several messages to api

lfalvarez:
This commit adds a config variable that will allow a user to make a writeit instance api read-only. Currently only works with Messages but it should probably help us fix the problem with sending several confirmation emails due to a bot. #623


lfalvarez:
I'm closing this as probably this isn't number one priority, and looking to gather more information on how to fix this

Issue #626, opened 2015-03-25
Email templates have string reprs rather than strings


duncanparkes:
Currently, the various email templates contain reprs of the strings that should be there rather than the strings themselves.

tmtmtmtm:
This means that emails are being sent with text that says things like:
  
```u'Please confirm your message```

rather than

```Please confirm your message```

PR #628, opened 2015-03-25
Unrepr the email templates.

duncanparkes:
Replace all the template strings with unrepr'd versions of themselves
where it doesn't cause an error.


tmtmtmtm:
Closes #626

PR #640, opened 2015-03-26
585 hide a message

lfalvarez:
This allows a user to 'hide' a message, in other words it turns the message non-public and fixes a {% url %} in messages_per_instance.html.


lfalvarez:
I don't really know =)

PR #644, opened 2015-03-27
After logout, redirect to '/'.

### LINKED ISSUE #633: When you sign out, go to WriteIt homepage
At the moment it goes to logged_out.html, and there's nothing we can do there that adds value at this stage

duncanparkes:
I've hard coded this to '/' rather than using reverse because
I can't see an easy way to get the reverse to use the right
subdomain without implementing a wrapper for the logout view
(which we could do in future).

Fixes #633.


duncanparkes:
This has been merged onto new_designs.

PR #673, opened 2015-03-30
Issues/610 resync popit

### LINKED ISSUE #610: Configuration of re-polling Popit(s)
Each associated PopIt can/will be re-polled at regular intervals, and your contacts updated accordingly.

Therefore we need to enable admins to:

- [ ] Turn auto-polling on/off
- [ ] Configure how often auto-polling happens
- [ ] Manually re-poll

There should not be a 'delete' concept for this behind the scenes, but we might want to simulate that in the front end.

lfalvarez:
This PR does:
* Adds a backend endpoint for manually syncing from popit
* Adds  ```periodicity``` for pulling from popit which can be set to: ```Never```, ```Once a day```, ```Twice a day``` and ```Once a week```. Being the default value *Once a week*.

In order for this to close #610, the missing parts are
* [ ] Adds the front end things for a user to change the values per instance.
* [ ] Adds a migration.

### DIFF
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -905,6 +905,13 @@ from tastypie.models import create_api_key
 
 models.signals.post_save.connect(create_api_key, sender=User)
 
+PERIODICITY = (
+    ('--', 'Never'),
+    ('2D', 'Twice every Day'),
+    ('1D', 'Daily'),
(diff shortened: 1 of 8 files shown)


lfalvarez:
Closing this in favor of #695

Issue #585, opened 2015-03-12
Admins should be able to 'unpublish' an email


tmtmtmtm:
See #582

martinszy:
I read a discussion about removing the private email functionality. The issue with that is tha the other option is to delete the message, and I'm not sure that everything I want unpublished I want deleted, I may want to keep record of things that I don't want published.
¿Or is part of the proposition that everything should be published?

tmtmtmtm:
Allowing a user to choose to send a message in private has only been removed from the web-based UI: the concept still exists in the core application. 

And, yes, there should absolutely be a way for admins to be able to 'hide' a message in problematic cases. Ideally this should be at the level of an individual message (or even just a single word/sentence/paragraph of it): not the entire conversation, but that's quite a lot of extra work, and may need to be a later addition. (Though, patches welcome!)

tmtmtmtm:
Closed by #674

PR #674, opened 2015-03-30
585 hide a message

lfalvarez:
This adds the ability for an admin to change the visibility of a message from  ```public``` to  ```hidden```.


tmtmtmtm:
Closes #585

PR #682, opened 2015-03-30
Design recipients screen

wrightmartin:
This makes ```manage/recipients/``` much easier to digest

### DIFF
--- a/nuntium/templates/nuntium/profiles/contacts/contacts-per-writeitinstance.html
+++ b/nuntium/templates/nuntium/profiles/contacts/contacts-per-writeitinstance.html
@@ -38,7 +38,7 @@ $(function() {
   <div class="page-header">
     <h2>{% trans "Recipients" %}</h2>
   </div>
-
+  <p>{% trans "There are" %} <strong>{{ writeitinstance.persons.count }}</strong> {% trans "recipients in your WriteIt" %}</p>
       {% autopaginate people %}
       {% paginate %}
       <table class="table table-striped recipients-table">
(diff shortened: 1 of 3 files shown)


wrightmartin:
screen shot 2015-03-30 at 16 28 03

nuntium/templates/nuntium/profiles/contacts/contacts-per-writeitinstance.html
```diff
@@ -38,24 +38,24 @@
   <div class="page-header">
     <h2>{% trans "Recipients" %}</h2>
   </div>
-
+  <p>{% trans "There are" %} <strong>{{ writeitinstance.persons.count }}</strong> {% trans "recipients in your WriteIt" %}</p>
```
tmtmtmtm: I'm assuming there's a better way to do this that will work in languages that form a sentence like this differently? (and to cope with the case of only having one person)

wrightmartin: @tmtmtmtm almost definitely, if a dev has two seconds I'm sure they could do this 1000% better

chrismytton: Yes, you can use [`{% blocktrans %}`](https://docs.djangoproject.com/en/1.6/topics/i18n/translation/#std:templatetag-blocktrans) to handle variables in language sentences and pluralization:

```
{% blocktrans with count person_count=writeitinstance.persons.count %}
There is 1 recipient in your WriteIt
{% plural %}
There are {{ person_count }} recipients in your WriteIt
{% endblocktrans %}
```

wrightmartin: amazing, thanks chris

chrismytton:
:+1:

PR #694, opened 2015-03-31
Fix tests after subdomains

lfalvarez:
This PR fixes several tests broken mostly due to urls.

This is done by overriding the behaviour of  ```client``` attribute of the TestCase by assigning the ```subdomain``` to the request.


chrismytton:
:+1: This looks good, going to merge this now and then fix the other tests once it's on `master`.

Issue #665, opened 2015-03-27
show nice status of popits


wrightmartin:
![screen shot 2015-03-27 at 17 17 59](https://cloud.githubusercontent.com/assets/2292925/6873026/3a8581f4-d4a5-11e4-84d3-fd76ec14e058.png)

Coloured blobs and that.

tmtmtmtm:
Is there a wireframe for how the configurable polling intervals will hook into this?

See also #627 and #610

wrightmartin:
@tmtmtmtm no, but I'll make one

wrightmartin:
- Is the polling interval the only thing we want to change?

![screen shot 2015-03-30 at 11 20 53](https://cloud.githubusercontent.com/assets/2292925/6894561/e2b24846-d6ce-11e4-8863-4d80e9242e42.png)

tmtmtmtm:
This is what appears when you click "Edit"?

From https://github.com/ciudadanointeligente/write-it/commit/1dbfcd742afea0887626edd45dbc0253e8364544 the current options for how often to re-poll are:
* Never
* Twice every Day
* Daily
* Weekly

The first of these would also take care of 'Disconnect' — and probably also 'Delete'.

We probably also need a manual "Repoll now"

wrightmartin:
Thanks, here's an updated wireframe

When you click _edit_ this slides out from the bottom with a little touch of JS. 

![screen shot 2015-03-30 at 12 39 42](https://cloud.githubusercontent.com/assets/2292925/6895618/0cade122-d6da-11e4-88a4-34d2ef960f13.png)

tmtmtmtm:
Looks good! :+1:

tmtmtmtm:
I'm wondering about having "Add a new data source" so prominent, though — in the majority of cases, there'll only ever be one source, so I'd be tempted to move this to the bottom, and/or make it also be a clicky slide-out thingy.

wrightmartin:
yeah, that sounds reasonable

tmtmtmtm:
Closed by #695

PR #695, opened 2015-03-31
Merge branches 610 665

### LINKED ISSUE #610: Configuration of re-polling Popit(s)

lfalvarez:
This PR does allow a user to manually resync from a popit instance and also to set a ```periodicity``` for repulling from popit. The possible values are  ```Never```,  ```Once a day```, ```Twice every day``` and ```Once a week```. The default value is ```Once a week```.

This closes #610 and #665 .


lfalvarez:
I think that there's still some work to be done in the frontend but I'm hoping to get feedback on this =)

PR #697, opened 2015-04-01
Issues/330 create an instance in frontend

lfalvarez:
This adds a link to create an instance in the landing page and also it provides the html for adding the instance.


chrismytton:
We'll need to open a separate ticket for styling this, but this bit looks good! :+1:

Issue #611, opened 2015-03-19
"Delete" a message should just be "Hide"


tmtmtmtm:
We shouldn't (under normal circumstances) ever be deleting messages (see also #593), so the "Delete" link in the Admin should just be "Hide" (presumably be re-toggling "Public")

lfalvarez:
oh! that's something close to a question I had in #200

chrismytton:
Fixed by https://github.com/ciudadanointeligente/write-it/pull/693

PR #698, opened 2015-04-01
This is just a proposal of changing deleting a message for hiding

lfalvarez:
I'm not particularly good a UX but as discussed in #611 probably what we need to have is just a hide button. This PR contains the hide funcionality wired up.

screepop


chrismytton:
Thanks! Merged into `design-messages-screen` as d1df7ac17f95652f0a9bf5549c6e4bbcbc626934.

PR #700, opened 2015-04-01
The email parser is will not complain if the email doesn't contain a UUI...

### LINKED ISSUE #699: When receiving an email without the unique identifier the email parser thinks it's an error
Sometimes emails are not a response to an actual message but spam. It would be good to store this emails but not to send an email if the parser could not determine to which message this response corresponds to.

lfalvarez:
...D telling write-it to which message is responding to. Closes #699.


chrismytton:
Looks good! :rocket:

PR #703, opened 2015-04-02
Design tweaks, mostly for message styling

zarino:
The "All messages" page, with an example of the sort of stuff we might put in the sidebar:

screen shot 2015-04-02 at 16 33 17

(The "Search" page, once it's done, could use this *exact same template*, since it's basically the same thing – a list of messages.)

The "Thread" page:

screen shot 2015-04-02 at 16 33 44

The "Preview message" page:

screen shot 2015-04-02 at 16 42 32

The homepage (that middle section still needs some work!! But it's out of scope for now):

screen shot 2015-04-02 at 16 36 29

### DIFF
--- a/nuntium/static/sass/instance/_layout-message.scss
+++ b/nuntium/static/sass/instance/_layout-message.scss
@@ -147,7 +147,8 @@
       position: absolute;
       font-size: 1.2em;
       left: 0.9em;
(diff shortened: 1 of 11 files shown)


chrismytton:
This is looking great! Thanks @zarino :+1:

Issue #677, opened 2015-03-30
Don't offer contacts for whom we have no email address


tmtmtmtm:
Later we can prompt users to find the contact info for us (#472); but for now simply don't even offer them at the writing stage.

PR #707, opened 2015-04-02
Issues/677 dont offer contacts for whom we dont have contacts

lfalvarez:
This PR filters the persons without contacts in the message creating part in the 'who' step.

### DIFF
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -58,6 +58,10 @@ class WriteItInstance(models.Model):
     def add_person(self, person):
         Membership.objects.get_or_create(writeitinstance=self, person=person)
 
+    @property
+    def persons_with_contacts(self):
+        return self.persons.filter(contact__writeitinstance=self, contact__isnull=False).distinct()
+
     def relate_with_persons_from_popit_api_instance(self, popit_api_instance):
         try:
             popit_api_instance.fetch_all_from_api(writeitinstance=self)
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -55,6 +55,9 @@ class WriteItInstance(models.Model):
         through='Membership')
     owner = models.ForeignKey(User, related_name="writeitinstances")
 
+    def add_person(self, person):
+        Membership.objects.get_or_create(writeitinstance=self, person=person)
+
     def relate_with_persons_from_popit_api_instance(self, popit_api_instance):
         try:
             popit_api_instance.fetch_all_from_api(writeitinstance=self)
--- a/nuntium/tests/message_creation_tests.py
+++ b/nuntium/tests/message_creation_tests.py
@@ -38,6 +38,9 @@ class MessageCreationTestCase(TestCase):
         self.assertIn(self.person1, who_form.fields['persons'].queryset)
         self.assertNotIn(self.person2, who_form.fields['persons'].queryset)
 
+        self.assertIn('writeitinstance', response.context)
+        self.assertEquals(response.context['writeitinstance'], self.writeitinstance)
+
     # def test_step_who_with_requestfactory(self):
     #     request = self.factory.get(url)
     #     add_session_to_request(request)
--- a/nuntium/tests/message_creation_tests.py
+++ b/nuntium/tests/message_creation_tests.py
@@ -1,17 +1,9 @@
 # coding=utf-8
 from global_test_case import GlobalTestCase as TestCase
-from nuntium.models import WriteItInstance
+from nuntium.models import WriteItInstance, Message
 from subdomains.utils import reverse
 from popit.models import Person
-from nuntium.forms import WhoForm
-
-# This method adds a session to a request
-# But I could not do anything with it
-# def add_session_to_request(request):
-#     """Annotate a request object with a session"""
-#     middleware = SessionMiddleware()
-#     middleware.process_request(request)
-#     request.session.save()
+from nuntium.forms import WhoForm, DraftForm
 
 
 class MessageCreationTestCase(TestCase):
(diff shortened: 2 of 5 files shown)


nuntium/models.py
```diff
@@ -55,6 +55,13 @@ class WriteItInstance(models.Model):
         through='Membership')
     owner = models.ForeignKey(User, related_name="writeitinstances")
 
+    def add_person(self, person):
+        Membership.objects.get_or_create(writeitinstance=self, person=person)
+
+    @property
+    def persons_with_contacts(self):
+        return self.persons.filter(contact__isnull=False)
```
duncanparkes: What happens if a Person has multiple Contacts? I think you're going to get a row per Contact/Person combination. Is that what you want?

nuntium/views.py
```diff
@@ -133,6 +132,7 @@ def get_context_data(self, form, **kwargs):
             context['persons'] = self.writeitinstance.persons.all()
         if self.steps.current == 'preview':
             context['message'] = self.get_form_values()
+            context['message']['people'] = context['message']['persons']
```
duncanparkes: I wonder here if we should just pick one of 'persons' and 'people' and stick with it.

nuntium/tests/message_creation_tests.py
```diff
@@ -0,0 +1,99 @@
...
+
+        response = self.client.get(url)
+        who_form = response.context['form']
+        self.assertIsInstance(who_form, WhoForm)
+        self.assertIn(self.person1, who_form.fields['persons'].queryset)
+        self.assertNotIn(self.person2, who_form.fields['persons'].queryset)
+
+        self.assertIn('writeitinstance', response.context)
+        self.assertEquals(response.context['writeitinstance'], self.writeitinstance)
+        recipient1 = self.writeitinstance.persons.first()
```
duncanparkes: If there are multiple people in `self.writeitinstance.persons`, I'd be more comfortable with choosing a particular one with `.get()`.

nuntium/tests/message_creation_tests.py
```diff
@@ -0,0 +1,99 @@
...
+        response = self.client.post(url, data={
+            'who-persons': [recipient1.id],
+            'write_message_view-current_step': 'who'
+            })
+        self.assertEquals(response.status_code, 302)
```
duncanparkes: This bit feels like a case for `assertRedirects`, which will handle the status code and the expected url in one go.

nuntium/tests/message_creation_tests.py
```diff
@@ -0,0 +1,99 @@
...
+        response = self.client.get(url2)
+        self.assertIsInstance(response.context['form'], DraftForm)
+        response = self.client.post(url2, data={
+            'draft-subject': 'This is the subjectand is very specific',
+            'draft-content': "This is the content",
+            'draft-author_name': "Feli",
+            'draft-author_email': 'email@mail.com',
+            'write_message_view-current_step': 'draft'
+            })
+        self.assertEquals(response.status_code, 302)
```
duncanparkes: Ditto here - `assertRedirects` is your friend.

nuntium/tests/message_creation_tests.py
```diff
@@ -0,0 +1,99 @@
...
+        url3 = reverse('write_message_step',
+            subdomain=self.writeitinstance.slug,
+            kwargs={'step': 'preview'})
+
+        self.assertEquals(response.url, url3)
+        response = self.client.get(url3)
+        self.assertEquals(response.context['message']['persons'][0].id, self.person1.id)
+        # in this step I'm oing to replicate the content of
```
duncanparkes: Typo s/oing/going/.

nuntium/tests/writeitinstances_test.py
```diff
@@ -141,6 +141,38 @@ def test_it_has_a_pulling_from_popit_status(self):
...
+        self.person2.contact_set.all()
+
+        self.assertTrue(self.person2.contact_set.all())
+        # self.person1 does not have contacts
+        # self.person2 has contacts
+        self.writeitinstance.add_person(self.person1)
+        self.writeitinstance.add_person(self.person2)
+
+        self.assertNotIn(self.person1,
+            self.writeitinstance.persons_with_contacts.all())
```
duncanparkes: You don't really need the `.all()` on these.

nuntium/views.py
```diff
@@ -90,8 +90,7 @@ def get_template_names(self):
 
     def get_form_kwargs(self, step):
         if step == 'who':
-            # FIXME: Do we actually want .all() here?
-            return {'persons_queryset': self.writeitinstance.persons.all()}
+            return {'persons_queryset': self.writeitinstance.persons_with_contacts.all()}
```
duncanparkes: You don't need the `.all()` here either - you've already done a filter.

nuntium/tests/message_creation_tests.py
```diff
@@ -0,0 +1,99 @@
...
+class MessageCreationTestCase(TestCase):
+    def setUp(self):
+        super(MessageCreationTestCase, self).setUp()
+        self.writeitinstance = WriteItInstance.objects.get(id=1)
+        self.person1 = Person.objects.get(id=1)
+        self.person2 = Person.objects.get(id=2)
+        self.writeitinstance.add_person(self.person1)
+        self.writeitinstance.add_person(self.person2)
+
+    def test_step_who_doesn_have_persons_without_contacts_self_client(self):
```
duncanparkes: I don't understand what this method name and comment are trying to say, I'm afraid.

duncanparkes:
Putting aside minor things, the only real issue is what happens if a person has more than one contact? Previously, you would have got a list with only one entry per person in it. I now think you'll get an entry per contact. I suspect this is not what you want, but I'm not sure. Perhaps pick the first email address for a person?

nuntium/models.py
lfalvarez: Ough! I hadn't seen that one! thanks!

lfalvarez: Actually this is a deeper thing, a person could have contacts related to another instance and therefore would be appearing in this queryset too! thanks this was a bug

nuntium/views.py
lfalvarez: I think that that the problem here is unify the naming related to ```Message.persons``` and ```Message.people``` for which I'm opening a new ticket.

chrismytton:
:rocket: :+1:

tmtmtmtm:
Closes #677

Issue #712, opened 2015-04-02
Hide or Replace "TODO: Show representatives here"


tmtmtmtm:
The homepage still has a "TODO: Show representatives here". If we don't have a "most popular representatives" function yet, simply hide this section.

Issue #711, opened 2015-04-02
Hide filters on /threads page


tmtmtmtm:
Until we can filter by All Messages | Awaiting Reply (#691), and/or have a public Search facility (#690), we shouldn't show these options.

Issue #706, opened 2015-04-02
Change public-facing test-mode banner


tmtmtmtm:
This currently says:

```
This instance is in Test Mode.

Nobody but you can access this instance, and any messages you send will be sent to your email address rather than real representatives
```

This is incorrect, as the instance can be seen by anyone who knows the URL.

Perhaps:

```
This site is in Test Mode.

Any messages you send will be sent to the site administrator rather than real representatives.
```

PR #744, opened 2015-04-07
Small frontend changes

lfalvarez:
The changes in this PR are:
* The text when displaying the test mode (#706) 
* Comments out the filters and search part in the ```/threads``` page (#711)
* Removing the *Showing most popular recipients* until we have something to show (#712)

### DIFF
--- a/nuntium/templates/base_instance.html
+++ b/nuntium/templates/base_instance.html
@@ -79,7 +79,7 @@
     <div class="site-alert">
         <div class="container">
             <h3 class="site-alert__heading">{% trans "This instance is in Test Mode." %}</h3>
-            <p class="site-alert__description">{% trans "Nobody but you can access this instance, and any messages you send will be sent to your email address rather than real representatives." %}</p>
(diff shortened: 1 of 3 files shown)


nuntium/templates/thread/to.html
```diff
@@ -4,9 +4,9 @@
 {% block content_inner %}
 
     <div class="person-profile">
-        <img class="person-profile__photo" src="http://placehold.it/128x128" width="128" height="128">
+        <img class="person-profile__photo" src="{% if person.image %}{{person.image}}{% else %}http://placehold.it/128x128{% endif %}" width="128" height="128">
```
tmtmtmtm: Showing the real image if it is exists is definitely an improvement(!), but I don't think that's a particularly good placeholder to show if there isn't one...

tmtmtmtm: If the person *does* have an image, do we really want to distort it to 128x128, regardless of the original dimensions?

If they don't, we want a better placeholder than this. Even http://upload.wikimedia.org/wikipedia/en/b/b1/Portrait_placeholder.png would be better.

lfalvarez:
I'm closing this in favour of #747, #748 and #749

PR #747, opened 2015-04-07
This changes the message for when an instance is in testing mode 

lfalvarez:
This changes the message for the testing mode.


duncanparkes:
:+1: I think we should accept this and worry about whether 'site administrator' is right later on.

tmtmtmtm:
@duncanparkes what's your worry with "site administrator"? (I agree we should fix it in another change, but it would be good to open an Issue for it!)

I took the opportunity to change 'This instance' to 'This site' as well, though.

lfalvarez:
:+1:

lfalvarez:
Is in master now =D

PR #748, opened 2015-04-07
Commenting out filtering and searching in the threads part.

lfalvarez:
This hides the filters and search form in the ```/threads``` page.


chrismytton:
Great! :+1:

PR #749, opened 2015-04-07
Removing the part for showing most popular recipients.

lfalvarez:
This removes the ```Showing most popular recipients``` part until further notice.

### DIFF
--- a/nuntium/templates/nuntium/instance_detail.html
+++ b/nuntium/templates/nuntium/instance_detail.html
@@ -24,10 +24,12 @@
                         {% include 'thread/message_mini.html' %}
                     {% endfor %}
                 </div>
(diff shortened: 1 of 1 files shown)


nuntium/templates/nuntium/instance_detail.html
```diff
@@ -24,10 +24,6 @@ <h2 class="instance-hero__heading">
                         {% include 'thread/message_mini.html' %}
                     {% endfor %}
                 </div>
-                <div class="instance-summary__popular">
```
duncanparkes: Use a Django `{% comment %}` so we remember where this should go when we put it back?

chrismytton:
:+1:

PR #759, opened 2015-04-08
Add a list of recipients to the draft screen

### LINKED ISSUE #648: /write/draft/ should show who the message is to
/en/write/draft/ should show the recipient(s) selected in the previous step

chrismytton:
This shows who you're writing to on the draft screen, which will become
important once we link directly to the draft screen from a person page.

Fixes #648

### DIFF
--- a/nuntium/templates/write/draft.html
+++ b/nuntium/templates/write/draft.html
@@ -1,5 +1,6 @@
 {% extends "base_instance.html" %}
 {% load i18n %}
+{% load nuntium_tags %}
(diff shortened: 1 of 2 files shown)


zarino:
I just used this screen myself a few minutes ago, and thought "wouldn't it be nice if I could see a (read-only) display of who I'm writing to" – it makes a big difference when you're writing a message, and sub-consciously you want to remind yourself who you're addressing and what tone you should adopt.

PR #766, opened 2015-04-08
Tweak wording of Test Mode banner

tmtmtmtm:
The doubling of "any messages you send will be sent" is a tiny bit jarring. Change it to "any messages you write will be sent".


chrismytton:
:+1:

Issue #775, opened 2015-04-08
Strip newlines from subject before sending notification


chrismytton:
When manually creating an answer the [subject should have it's `\n` stripped before sending](https://github.com/ciudadanointeligente/write-it/blob/d96859606b46e795696193d1d79416428ab1c366/nuntium/models.py#L489), otherwise the following error occurs:

```
Internal Server Error: /en/manage/messages/6/answers/create/
Traceback (most recent call last):
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 29, in _wrapper
    return bound_func(*args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/utils/decorators.py", line 25, in bound_func
    return func(self, *args2, **kwargs2)
  File "/vagrant/nuntium/user_section/views.py", line 281, in dispatch
    return super(AnswerEditMixin, self).dispatch(*args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/views/generic/base.py", line 87, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/views/generic/edit.py", line 205, in post
    return super(BaseCreateView, self).post(request, *args, **kwargs)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/views/generic/edit.py", line 171, in post
    return self.form_valid(form)
  File "/home/vagrant/writeit-virtualenv/local/lib/python2.7/site-packages/django/views/generic/edit.py", line 147, in form_valid
    self.object = form.save()
  File "/vagrant/nuntium/user_section/forms.py", line 147, in save
    answer.save()
[...]

PR #778, opened 2015-04-08
Strip newlines from subject before sending

chrismytton:
Because the subject line is read from a file it contains a newline, we
need to strip this otherwise the email library will throw an error.

Fixes https://github.com/ciudadanointeligente/write-it/issues/775

### DIFF
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -512,7 +512,7 @@ def send_new_answer_payload(sender, instance, created, **kwargs):
 
         for subscriber in subscribers:
             msg = EmailMultiAlternatives(
-                subject,
+                subject.strip(),
                 text_content,
                 from_email,
                 [subscriber.email],


lfalvarez:
:+1:

lfalvarez:
are there any other places where this could turn into an error?

tmtmtmtm:
Merging this one for now, though it would be good to add a test that catches this.

lfalvarez:
I think that these are some of the places where this could happen:

* https://github.com/ciudadanointeligente/write-it/blob/master/nuntium/models.py#L423
* https://github.com/ciudadanointeligente/write-it/blob/master/nuntium/models.py#L795
* https://github.com/ciudadanointeligente/write-it/blob/master/mailit/__init__.py#L68

tmtmtmtm:
@lfalvarez feel free to fix the others, and add tests for them (#786)

PR #793, opened 2015-04-08
Thread boxes depend on public search, not Admin

tmtmtmtm:
The comment here points at the wrong Issue: we can put this back when we have public search, which is different from the admin side filtering.


chrismytton:
Looks good :+1:

PR #797, opened 2015-04-08
Hide a message rather than deleting it when moderation has been rejected

### LINKED ISSUE #593: Rejecting a message should not delete it
In the normal case, we shouldn't really be deleting any messages.

lfalvarez:
Closes #593

### DIFF
--- a/nuntium/views.py
+++ b/nuntium/views.py
@@ -166,7 +166,12 @@ class RejectModerationView(ModerationView):
 
     def get(self, *args, **kwargs):
         get = super(RejectModerationView, self).get(*args, **kwargs)
-        self.get_object().message.delete()
+        self.object.message.public = False
+        # It is turned True to avoid users to
+        # mistakenly moderate this message
+        # in the admin section
+        self.object.message.moderated = True
+        self.object.message.save()
         return get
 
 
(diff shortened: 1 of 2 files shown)


nuntium/views.py
```diff
@@ -166,7 +166,8 @@ class RejectModerationView(ModerationView):
 
     def get(self, *args, **kwargs):
         get = super(RejectModerationView, self).get(*args, **kwargs)
-        self.get_object().message.delete()
+        self.object.message.public = False
```
chrismytton: I think you'll also need to set `self.object.message.moderated = True` here as well. Otherwise the **Moderated** column in the list of messages in the manage section will say ["This message has not been moderated and you can moderate it by clicking here"](https://github.com/ciudadanointeligente/write-it/blob/cdf6ec79c65467f58a337b82154b4249911f1f3c/nuntium/templates/nuntium/profiles/messages_per_instance.html#L81-L93), and when a user clicks it they might inadvertently accept the message.

lfalvarez: Wow! thanks I didn't see that one coming!

chrismytton:
:+1: Looks good!

PR #798, opened 2015-04-08
Commenting out the report button until we know the best way to proceed. ...

### LINKED ISSUE #713: "Report for Abuse" button doesn't work
A thread page currently has a "Report for Abuse" link that doesn't go anywhere. 

We should either make this a mailto: (per #200), or hide it until it does something.

lfalvarez:
Fixes #713


chrismytton:
Looks good! :+1:

Issue #714, opened 2015-04-02
Design for pagination of /threads


tmtmtmtm:
http://nouabook.writeit.ciudadanointeligente.org/en/threads/, for example, has a lot of messages and is very slow.

tmtmtmtm:
The pagination has been merged — this still needs design applied as per [comment](https://github.com/ciudadanointeligente/write-it/pull/803#issuecomment-90958350) from @zarino

zarino:
Pull request: https://github.com/ciudadanointeligente/write-it/pull/924

PR #803, opened 2015-04-08
Paginate the list of all messages for an instance

chrismytton:
This will probably need a bit of design from @zarino 

screen shot 2015-04-08 at 16 47 09

Part of https://github.com/ciudadanointeligente/write-it/issues/714 (but still needs design)


zarino:
@chrismytton: Assuming you uncomment the `// @import "../bootstrap/pagination";` line in `instance/_bootstrap.scss`, you'll be able to get most of the way by just using the markup from this Bootstrap example:

http://getbootstrap.com/components/#pagination

PR #807, opened 2015-04-08
Lots of fixes for the person page

### LINKED ISSUE #639: Button to "write to this person" on the person profile page
Where "person" is an MP or a representative or whatever.

This page is likely to be picked up by search engines, so we should make sure visitors who land on it are able to quickly start a new message to the person.

Ideally, clicking the button should take you to the "Select recipients" stage of the writing process, with the person's name pre-filled into the text input. @chrismytton @duncanparkes does that sound feasible?

### LINKED ISSUE #754: Page listing all messages from a person: Sort out content and layout
`nuntium/templates/threads/to.html` currently looks like:

screen shot 2015-04-07 at 16 35 49

As you can tell from the lack of contrast, and weird spacing, it needs attention.

Issues of particular note:

* Needs a "Write to this person" button (#639) maybe with some supporting text explaining the concept very quickly for visitors who’ve arrived via a web search, for example?
* Most recipients won’t have a summary, so the "header" (if we keep it) should work even when we know only the person's name.
* The messages should be of the `thread/message-mini.html` style, not the full message.
* The messages should be narrower on a desktop. Maybe switch to a two-column view, with person info and "Write to this person" bits in the left column, and messages down the right?
* The messages should be on a light grey background, to make them stand out, as they are everywhere else. This might necessitate the rest of the content being on the same background too.

Probably best *not* to address all of these issues in a single pull request, unless it can be turned around quickly.

zarino:
Now looks like:

screen shot 2015-04-08 at 17 09 52

Fixes #754. Also fixes #639. And improves the styling of the "To" field introduced in #759 since I was in the vacinity.

### DIFF
--- a/nuntium/static/sass/instance/_layout-person.scss
+++ b/nuntium/static/sass/instance/_layout-person.scss
@@ -24,5 +24,9 @@
 
   .results__actions {
     @extend .col-sm-4;
+
+    .btn-primary {
+      margin-top: 1em;
+    }
   }
 }
--- a/nuntium/static/sass/instance/_layout-person.scss
+++ b/nuntium/static/sass/instance/_layout-person.scss
@@ -1,24 +1,28 @@
 .person-profile {
-  padding: 1.5em 0;
-  border-bottom: 1px solid $gray-lighter;
-  overflow: auto;
+  @include clearfix();
+  margin-top: 1em;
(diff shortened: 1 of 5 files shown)


chrismytton:
Looks good! :+1:

PR #812, opened 2015-04-08
Issues/724 adding a source should let you know if has already been linked

lfalvarez:
This lets you know that has been an error, when you try to relate a popit instance to a WriteIt that already has that popit instance related.

Here it is how it looks.

foto

It would probably need a bit of styling and wording any suggestions @tmtmtmtm @zarino ?

### DIFF
--- a/nuntium/forms.py
+++ b/nuntium/forms.py
@@ -138,17 +138,7 @@ popit_urls_completer = [
 ]
 
 
-class WriteItInstanceCreateFormPopitUrl(ModelForm):
-    popit_url = URLField(
-        label=_('Url of the popit instance api'),
-        help_text=_("Example: http://popit.master.ciudadanointeligente.org/api/"),
-        required=False,
-        )
-
-    class Meta:
-        model = WriteItInstance
-        fields = ('owner', 'name', 'popit_url')
-
+class PopitParsingFormMixin(object):
     def other_possible_popit_url_parsings(self, popit_url):
         if popit_url.startswith('https://'):
             popit_url = popit_url.replace('https://', 'http://', 1)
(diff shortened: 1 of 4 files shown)


lfalvarez:
It actually says something like ```You already have this popit instance related```

tmtmtmtm:
Maybe "This PopIt was already previously added"?

zarino:
Or "You are already pulling data from this PopIt"

Followed by… (feature creep!)… "But we are updating the database anyway, to pull in any new contacts" – since I have a feeling that, when people add a source, they probably "just want to get all the people from this popit into this writeit", and telling them the popit is already associated doesn't help them get people in.

(Feel free to throw the last paragraph out if there's something factually incorrect about how I assume the popit/writeit integration works)

tmtmtmtm:
I started thinking through things like that (including whether we actually need to show an error at all, or just imply success, as the PopIt that they asked to be linked has been!) but quickly backed away again, as it seemed like a can of worms.

But if it's trivial to treat a re-adding of an existing popit as a request to re-fetch it, I'm all for doing that. 

@lfalvarez: if this is something that could be done in under 15 minutes, feel free to do it here. Otherwise let's just tweak the wording, and open a separate ticket for doing something better.

lfalvarez:
So it isn't a very complex thing to do, but I'm wondering: Are we updating from a popit when someone said please create me a new one? if the answer to that is yes, please feel free to accept #826 that does exactly that.

tmtmtmtm:
I've tweaked the wording in #826 — I suspect it should be merged into this Pull Request before doing anything further here!

chrismytton:
Looks good! :+1:

Issue #808, opened 2015-04-08
Stats: No need for All Messages and Public Messages unless they differ


tmtmtmtm:
![stats 2015-04-08 at 17 27 50](https://cloud.githubusercontent.com/assets/57483/7049887/283fae56-de15-11e4-96df-191f0afe799e.png)

Now that the stats page doesn't show how many Private Messages there are (unless there are any), there's no need to show both "Total messages" and "Total public messages".

PR #814, opened 2015-04-09
Only show total public messages if there are private ones

chrismytton:
On the stats page it doesn't make sense to show a count of public
messages and total messages if there are no private messages and
therefore the counts are the same.

Fixes https://github.com/ciudadanointeligente/write-it/issues/808


duncanparkes:
:+1:

Issue #810, opened 2015-04-08
Key Error when creating a New Message


tmtmtmtm:
Since #790 has gone live, I'm now getting a KeyError when trying to create a new message:

![key error 2015-04-08 at 17 38 07](https://cloud.githubusercontent.com/assets/57483/7050020/1060f87a-de16-11e4-905d-a302e391070d.png)

lfalvarez:
I believe it has to do with the lines https://github.com/ciudadanointeligente/write-it/blob/master/nuntium/views.py#L130-L133 and with that probably https://github.com/ciudadanointeligente/write-it/blob/master/nuntium/forms.py#L101-L110 doesn't have a ```persons``` within its fields

PR #818, opened 2015-04-09
Fix KeyError when creating a new message

chrismytton:
Fixes https://github.com/ciudadanointeligente/write-it/issues/810

### DIFF
--- a/nuntium/views.py
+++ b/nuntium/views.py
@@ -129,7 +129,7 @@ class WriteMessageView(NamedUrlSessionWizardView):
             context['persons'] = self.writeitinstance.persons.all()
         if self.steps.current == 'preview' or self.steps.current == 'draft':
             context['message'] = self.get_form_values()
-            context['message']['people'] = context['message']['persons']
+            context['message']['people'] = context['message'].get('persons')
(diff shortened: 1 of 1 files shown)


nuntium/views.py
```diff
@@ -83,6 +83,18 @@ def dispatch(self, request, *args, **kwargs):
         self.writeitinstance = get_object_or_404(WriteItInstance, slug=request.subdomain)
         return super(WriteMessageView, self).dispatch(request=request, *args, **kwargs)
 
+    def get(self, *args, **kwargs):
+        # Check that the form contains valid data
```
lfalvarez: That's a good one! I'm thinking that perhaps we could separate this function into a different sort of ```is_valid``` thing that could perhaps do some other validations like for example rate limiting, but that could obviously go in another issue. I think that the idea would be to just hook validators into the form.

lfalvarez:
:+1: good to go!

nuntium/views.py
tmtmtmtm: @lfalvarez that's a good idea — can you open an issue for doing that?

PR #819, opened 2015-04-09
Issues/773/incomming email replies cut oddly

### LINKED ISSUE #773: Incoming Replies cut oddly
response 2015-04-08 at 12 19 34

This includes stuff that shouldn't be there (or else is missing the rest of the quoted message)

http://eduskunta-2.writeit.ciudadanointeligente.org/en/thread/will-you-be-supporting-bill-19481-9f-2/

lfalvarez:
This eliminates odd parts of the email by removing entirely the "To" header from the content of the email.

So for example if the email came to: ```Felipe <instance1235+key@writeinpublic.org>``` (this is a response email). And the email contains a signature like ```Il giorno 25/mar/2015, alle ore 16:56, Felipe 
<instance1235+key@writeinpublic.org> ha scritto:```. The output will be ```Il giorno 25/mar/2015, alle ore 16:56,   ha scritto:```.

Sometimes in the email content the signature will be cut, for example: 

```
Tony
<the-instance+identifier@writeinpublic.org>
```
This will also remove that signature.

The possible edge case could be when the newlines come between the names for example:
```
To: Tony <the-instance+identifier@writeinpublic.org>
```
and the content goes like: 
```
T
o
ny 
<the-instance+identifier@writeinpublic.org>
```
If that's the case this will not catch that one.

Fixes #773

### DIFF
--- a/mailit/bin/handleemail.py
+++ b/mailit/bin/handleemail.py
@@ -1,3 +1,4 @@
+# coding=utf8
 import email
 import re
 import requests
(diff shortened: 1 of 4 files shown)


chrismytton:
@lfalvarez I'm struggling to do a manual test on this locally, does this handle removing the date from the bottom of the message as well? If not then it should as it doesn't make sense to have the date in the content of the message as well as in the headers.

lfalvarez:
oh for the manual test you would have to use an example email like [this one](https://github.com/ciudadanointeligente/write-it/blob/issues/773/incomming-email-replies-cut-oddly/mailit/tests/fixture/mail_from_tony.txt)  turn it into a txt file and change the identifier ```4aaaabbb``` for one that actually exists and then do this in the terminal ```./manage.py handleemail mail_from_tony.txt``` that could actually work as well for parsing attachments.
And as for the dates I think it is quite hard specially because they can come in different languages and formats.

chrismytton:
I think *eventually* it would be good to try and do something to get rid of the date as well, but for now this is certainly an improvement over what we currently have. :+1:

PR #838, opened 2015-04-10
Site Manager: Add link to Recipients page

### LINKED ISSUE #831: Site Manager: Add link to Recipients page
your sites 2015-04-10 at 09 45 11

Each site currently gives a link to the Messages page — it should similarly link directly to the Recipients page (again, with a similar count of how many there are). Otherwise you need to go to "Edit", just to see who they are, which is slightly confusing. (Plus we have spare room here, now that we've removed the Popit API URL.)

[Driveby: rename "PopIts" to "Sites"? That'll be happening anyway in #530, but we should do some of it in passing to make that process easier.]

tmtmtmtm:
Add a new 'Recipients' column to the Site Manager:

your sites 2015-04-10 at 13 03 02

Closes #831


chrismytton:
Looks good! :roc

PR #841, opened 2015-04-10
Styling for "future replies" message on thread page

zarino:
I noticed [this](https://github.com/ciudadanointeligente/write-it/blob/650bc976141d480d74fb53361be9dd2c54fdc93a/nuntium/templates/thread/read.html#L39) crept in over the last two days, and looked a bit rushed.

This change styles the message so that it looks more like a placeholder for a reply.

screen shot 2015-04-10 at 13 14 30


tmtmtmtm:
Much better! :rocket:

PR #842, opened 2015-04-10
Disallow negative rate_limits

### LINKED ISSUE #840: Admin: Rate Limit shouldn't allow negative numbers
I'm not entirely sure what a Rate Limit of -1 would be, but we should probably disallow it (unless it has a useful meaning that I can't think of)

negative rate 2015-04-10 at 13 13 26

tmtmtmtm:
rate_limit should always be above zero.

I’ve added the server-side validation for this, but can’t work out yet how to add an equivalent client-side check to the one on maximum_recipients.

Fixes #840

### DIFF
--- a/nuntium/templates/nuntium/writeitinstance_update_form.html
+++ b/nuntium/templates/nuntium/writeitinstance_update_form.html
@@ -52,6 +52,7 @@ $(".chosen-person-select").chosen();
           {{advanced_form.allow_messages_using_form}}
       </div>
       <div class="form-group">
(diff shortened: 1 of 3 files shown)


tmtmtmtm:
I suspect the *correct* fix for this is to change rate_limit  in the model to a [PositiveIntegerField](https://docs.djangoproject.com/en/1.7/ref/models/fields/#positiveintegerfield), but that's not an area I really want to go meddling with, so I suggest we take this route for now...

lfalvarez:
it looks pretty good to me!

PR #843, opened 2015-04-10
Make all Message fields compulsory

### LINKED ISSUE #820: Message Composer: Add some client-side validation
Currently, checking that you've filled in your name / email address etc looks like it's only doing server-side validation. I'm presuming it's fairly trivial to add a basic client-side check too?

tmtmtmtm:
Add `’required’: True` to all the widgets of the Message writing form
so that we get HTML5 client-side validation, as well as the server-side
validation that already exists.

required message 2015-04-10 at 14 46 48

Closes #820


lfalvarez:
This looks pretty good :+1:

PR #848, opened 2015-04-10
728: Explain import errors

### LINKED ISSUE #728: Sources manager: Explain import errors
As a user I should be told *why* my import has errored so I can try to rectify the situation.

Currently if an import does error we're [setting a `status_explanation` property](https://github.com/ciudadanointeligente/write-it/blob/156d70056f123f6d1b4056ebe24ea2d1dab3c107/nuntium/models.py#L947-L950) but we're not actually displaying it anywhere. Perhaps this could be a tooltip on the error or something?

screen_shot_2015-04-07_at_11 31 16

tmtmtmtm:
If we have a status_explanation for the polling status, show it:

source error 2015-04-10 at 15 42 26

Fixes #728


chrismytton:
Looks good! :+1:

Issue #827, opened 2015-04-09
Message Manager is still confusing


tmtmtmtm:
![message manager 2015-04-09 at 22 16 02](https://cloud.githubusercontent.com/assets/57483/7077139/6a0afbb4-df06-11e4-8391-ac942f5b8c4a.png)

We have four rows that show various binary options for a message, with an unrelated row in the middle that tells you how many responses it has had. 

Two of them are purely informational as to the status (though at least one of those _should_ let you change it #782); two of them _do_ let you change it — but in different ways. 

Two of them (at opposite ends! #720) refer to to the same status (public/private) — one just telling you what it is, the other providing a somewhat confusing icon that lets you toggle the status (though it probably shouldn't #779.)

"Moderate" (as a text button this time, and misleading text at that, see #583) only lets you set to "Yes" — there's no way to reject a message from here — and once clicked, turns into a purely informational tick/cross combo. Furthermore, if messages exist from before Moderation was even turned on in the instance (as here), clicking "Moderate" on those earlier messages will simply give an error (#704). The "Moderate" button also displays for messages that haven't been confirmed yet — clicking it gives a 'The message needs to be confirmated first" error.
 
Some suggestions (though this might just need an entire rethink):
- [x] Separate the "number of replies" from the other columns (at least move it to the leftmost column, but preferably do something entirely better)
- [x] Don't show the "Moderate" button for messages it doesn't apply to (#704)
- [x] Have separate "Approve / Reject" actions for messages needing moderation (#583)
- [x] Group the two "Public?" columns together (#720)
- [ ] Add a confirmation step to making a message public that is currently private (#779) 
- [x] Shouldn't say "in your WriteIt" [#530] — probably best to just drop this entirely here

tmtmtmtm:
See also #858

tmtmtmtm:
Everything here is either done, or ticketed separately, so closing this one.

PR #849, opened 2015-04-10
583: Improvements to Message Manager

### LINKED ISSUE #583: Change misleading use of "moderate"
This seems like clicking the "Moderate" button will take you to somewhere that you perform the moderation — not that clicking the button will Approve the message. (See #582)

moderate it 2015-04-09 at 21 46 39

Possibly use "confirm" instead.

tmtmtmtm:
Make it more obvious that the button actively approves the message. (Closes #583)

Whilst here, add ?s to the text headers (`Moderated?` instead of `Moderated`), move the "number of replies" column further left, and remove needless `WriteIt` name. (Improves #827)

message manager 2015-04-10 at 16 32 38

### DIFF
--- a/nuntium/templates/nuntium/profiles/messages_per_instance.html
+++ b/nuntium/templates/nuntium/profiles/messages_per_instance.html
@@ -26,7 +26,7 @@
     {% blocktrans count message_count=writeitinstance.message_set.count %}
     <strong>1</strong> message in your WriteIt
     {% plural %}
-    There are <strong>{{ message_count }}</strong> messages in your WriteIt
+    There are <strong>{{ message_count }}</strong> messages 
     {% endblocktrans %}
   </div>
     {% autopaginate message_list %}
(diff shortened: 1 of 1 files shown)


nuntium/templates/nuntium/profiles/messages_per_instance.html
```diff
@@ -26,7 +26,7 @@
     {% blocktrans count message_count=writeitinstance.message_set.count %}
     <strong>1</strong> message in your WriteIt
```
chrismytton: Should probably remove "in your WriteIt" here as well.

chrismytton:
One minor comment, but otherwise looks good :+1:

nuntium/templates/nuntium/profiles/messages_per_instance.html
tmtmtmtm: The screenshot makes me think that I left this out because another change already did that … but I can't find that right now

tmtmtmtm: Ah, Ignore that — Just worked out that I got the plural but not the singular. Yep, should definitely fix that.

PR #854, opened 2015-04-10
Source Manager: Give each form's div a unique id

### LINKED ISSUE #851: Source Manager: Can only edit first Source
If have multiple sources, I get multiple "Edit" links:

data sources before 2015-04-10 at 16 48 39

But no matter which of the three I click, it's the top source that gets an editor:

data sources after 2015-04-10 at 16 48 49

tmtmtmtm:
Currently every form in the sources list gets the same ID, so attempting to edit any of them will always give you the first one.

Instead give each a unique id, so that we can edit any of them:

edit second data source 2015-04-10 at 17 11 56

Fixes #851


chrismytton:
Looks good :+1:

PR #855, opened 2015-04-10
Source Manager: Fix confusing buttons

### LINKED ISSUE #722: Source Manager: Ungroup "Fetch Data Now" and "Save"
When editing a Data Source, the "Fetch New Data Now" and "Save" buttons look as if they're both connected to setting the Polling Interval:

source manager 2015-04-07 at 11 10

But the "Fetch Now" is an entirely separate concept, and shouldn't appear in the middle of changing the Interval.

tmtmtmtm:
**NOTE**: This should be merged after 854, as I mistakenly branched off it.

Having the “Fetch new data now” button in-between changing and saving the polling interval is confusing. Move the ‘Save’ button up to beside the interval:

data source 2015-04-10 at 17 21 47

Fixes #722


chrismytton:
Looks good :+1:

PR #859, opened 2015-04-10
Make sure user is logged in before moderation

### LINKED ISSUE #834: Moderation: Links in email don't require being logged in
If I click on the Accept/Reject links in a "Message requires moderation" email, I don't need to be logged in for them to take effect.

I haven't tested yet to see if this is true for other such links, but we should make sure that all admin functionality only works if you're actually signed in as the correct person!

chrismytton:
Fixes #834


nuntium/user_section/views.py
```diff
@@ -331,13 +319,68 @@ def get_redirect_url(self, *args, **kwargs):
...
+
+    def get_redirect_url(self, *args, **kwargs):
+        message = get_object_or_404(Message,
+            pk=kwargs['pk'],
+            writeitinstance__slug=self.request.subdomain,
+            writeitinstance__owner=self.request.user
+            )
+        message.public = False
+        message.moderated = True
+        message.save()
```
lfalvarez: I'm thinking that perhaps we could do something with the OutboundMessage related to this message, apparently they should keep on being in status 'new' as [in here](https://github.com/ciudadanointeligente/write-it/pull/859/files#diff-5e559ebeaac71a53dc393dcd11f56404R398), but I think that since we're rejecting the message we could may be set those OutboundMessages to 'error' or something. Or even we could create a new status for them 'rejected'. I'd just be afraid that some of those messages keep around and they could still be sent to the recipient.

PR #865, opened 2015-04-11
Tidy the "Create a new site" forms

### LINKED ISSUE #863: Sites Manager: Improve the New Site form
Appears as a Modal at /accounts/your_instances

new site modal 2015-04-11 at 18 35 11

Or a page at http://writeit.ciudadanointeligente.org/en/writeitinstance/create

new site page 2015-04-11 at 18 37 19

- [ ] Give an example URL that resolves
- [ ] We don't need a "Close" button if we're in a Modal
- [ ] "Submit" could be "Create" instead

Split out to ticket #877 instead:
- Default to only asking for the URL (and fill the name in automagically)
- Explain better the connection with PopIt

tmtmtmtm:
Closes #863 

- Make the name and URL both be Required 
- We don't need API URLs — simple PopIt URLs will do
- Use working PopIts as examples
- Don't have a separate Modal form from Sites Manager: just link to the main one
- Use the "Site" terminology instead of WriteIt/Instance

Splitting the form into a two-step process that fetches the Popit's /about page to auto-fill the `name` field doesn't work yet, because CORS. So that half is now #877

### DIFF
--- a/nuntium/forms.py
+++ b/nuntium/forms.py
@@ -160,9 +160,9 @@ class PopitParsingFormMixin(object):
 
 class WriteItInstanceCreateFormPopitUrl(ModelForm, PopitParsingFormMixin):
     popit_url = URLField(
-        label=_('Url of the popit instance api'),
-        help_text=_("Example: http://popit.master.ciudadanointeligente.org/api/"),
-        required=False,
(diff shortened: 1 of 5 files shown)


chrismytton:
Looks good :+1:

PR #869, opened 2015-04-11
Show update frequency

### LINKED ISSUE #853: Source Manager: Show polling intervals
Currently you can't see from the Source Manager when a PopIt will be re-polled:

data sources 2015-04-10 at 17 07 24

You can only get the information by switching to Edit mode:

edit data sources 2015-04-10 at 17 07 51

tmtmtmtm:
In the Source Manager, show the update frequency without needing to click the ‘Edit’ button. Closes #853

update frequency 2015-04-11 at 22 10 54


chrismytton:
Looks good :+1:

PR #870, opened 2015-04-11
Add the instance_slug to the hidden 'Who'

### LINKED ISSUE #856: Write to this person button broken since adding prefix to write wizard
e.g. http://parliament-of-finland.writeit.ciudadanointeligente.org/en/to/19530/

This button is broken because we've changed the prefix, so we need to update the hidden fields in that form.

tmtmtmtm:
The form prefixes we added in #790 broke the hidden form we use to “Write to this person” on the Recipient page. We need to add the instance_slug to make it work again. Fixes #856


chrismytton:
Looks good :+1:

PR #873, opened 2015-04-11
Message Creator: Order contacts by name

### LINKED ISSUE #872: Message Creator: Recipient list is unsorted
recipient search 2015-04-11 at 23 57 19

Even though we're doing a search, it would be good to sort the list.

tmtmtmtm:
In the “Search for recipients by name” box, order them by name.

Before:

recipient search 2015-04-11 at 23 57 19

After:

sorted recipient search 2015-04-12 at 00 08 59

Ideally we’d be able to order by them by their PopIt sort_name, but this is better than not ordering them at all.

Closes #872

### DIFF
--- a/nuntium/views.py
+++ b/nuntium/views.py
@@ -100,7 +100,7 @@ class WriteMessageView(NamedUrlSessionWizardView):
 
     def get_form_kwargs(self, step):
         if step == 'who':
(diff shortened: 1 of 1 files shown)


chrismytton:
Looks good :+1:

Issue #752, opened 2015-04-07
Recipient Manager should filter by Active or Not


tmtmtmtm:
If you connect a large PopIt with lots of historic data, the Recipient Manager is quite unwieldy:

![recipient manager 2015-04-07 at 15 50 56](https://cloud.githubusercontent.com/assets/57483/7025824/f5c5ee72-dd3d-11e4-9cb5-690517507d97.png)

We should offer views into this by any of the mechanisms we use for automatically deciding who should be active, particularly

- [ ] All currently active recipients
- [ ] People who *would* be active if we had an email address for them

PR #875, opened 2015-04-11
Recipient Manager: Order contacts by name

### LINKED ISSUE #874: Recipient Manager: Order by name
The Recipients Manager is completely unsorted, making it almost impossible to find a specific person if your instance is large.

recipients manager 2015-04-12 at 00 12 00

tmtmtmtm:
Ideally this would be by their `sort_name`, or, if we don’t have that, we should really take into account the language code of the linked PopIt. But this will do for now, at least until #752 gives us filtering options.

Before:

recipients manager 2015-04-12 at 00 12 00

After:

sorted recipient manager 2015-04-12 at 00 23 54

Closes #874


chrismytton:
Looks good :+1:

PR #878, opened 2015-04-12
Show the maximum recipients on the form

### LINKED ISSUE #857: Message Creation: When you hit the max_recipients limit, provide an explanation
Currently users are restricted in how many people they can write to. But when they hit that limit, it just silently stops letting them add more recipients. Instead it should let them know what's happening.

hit max recipients 2015-04-12 at 02 25 24

tmtmtmtm:
When searching for recipients, show how many you can have. Closes #857 

inform max recipients 2015-04-12 at 02 24 38


chrismytton:
Looks good :+1:

Issue #751, opened 2015-04-07
Site Manager: Admins shouldn't be able to delete live sites 


tmtmtmtm:
Once a site has been put live (no longer in test mode), admins shouldn't be able to simply delete it.

Instead they should simply be able to prevent new messages being created.

I think the underlying parts required for this are all in place: it's mostly just a matter of changing /manage/delete from

![delete instance 2015-04-07 at 15 45 42](https://cloud.githubusercontent.com/assets/57483/7025730/71847b24-dd3d-11e4-808c-2efca4eab23c.png)

to give a button with the same effect as disabling "Allow messages using form" on the Basic Update page.

tmtmtmtm:
#880 has removed this interface for this — but we should really disallow it at a deeper level.

PR #881, opened 2015-04-12
Split live sites and test sites

### LINKED ISSUE #880: Site Manager: Separate Live Sites and Test Sites
Show the live sites and test sites separately in the Site Manager. Then, as a step towards #751, we can only provide a delete button for test sites. 

We can also give a good explanation as to what Test Sites mean — this will be especially useful for new users: their first experience will be of having a single site, in Test Mode.

tmtmtmtm:
Show the live sites and test sites separately in the Site Manager.

Before:

grouped sites 2015-04-12 at 03 39 09

After:

split sites 2015-04-12 at 03 22 31

Only test sites now have a Delete button (see #751)

It’s important for the Test Sites to get some better explanatory text, as the New User journey will involve them only having a single site, in test mode. The current text is just a placeholder that will remind us to create something better.

NB: I don't really like how much duplication there is in the template, and I’m not entirely convinced that the views.py implementation of this is the right way to go about this at all... It's also a little ugly, but the admin design process should hopefully clean that up.

Closes #880

### DIFF
--- a/nuntium/templates/nuntium/profiles/your-instances.html
+++ b/nuntium/templates/nuntium/profiles/your-instances.html
@@ -39,20 +39,63 @@
   </div>
 {% endif %}
 
-    <div class="page-header">
-    <h2>{% trans "Your Sites" %}</h2>
-    </div>
+      <div class="page-header">
+        <h2>{% trans "Live Sites" %}</h2>
+      </div>
 
-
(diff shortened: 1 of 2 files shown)


chrismytton:
There's probably a way to clear up some of the duplication with partials etc, but probably not worth it if the design is changing anyway, so… :+1:

PR #884, opened 2015-04-12
Move the Test Mode banner to the top

### LINKED ISSUE #882: Test Mode status bar should always be at the top
delete instance 2015-04-12 at 03 12 46

The banner for Polling Information shouldn't appear above the Test Mode banner. That one should always be at the very top.

tmtmtmtm:
The “Test Mode” banner should always be above other status messages.

Also, tweak the wording to be more in line with the public-facing
equivalent of it.

reordered banners 2015-04-12 at 10 46 05

Closes #882


chrismytton:
:+1: Looks good

PR #885, opened 2015-04-12
Remove double flash after site creation

### LINKED ISSUE #839: Site Manager: Double flash message when creating a new Site
When creating a new site, two flash messages are given: 

double flash 2015-04-10 at 13 05 59

At the top we say "Currently we are pulling people from http://eduskunta.popit.mysociety.org/api/v0.1"

and in the Editing section we say: "We are getting information from popit about the people and contacts in this instance"

These should be combined, and expanded to give more information about why we're now dropping them into a scary Admin interface. 

Or perhaps we should have an entirely new "New Site Created page" that eases them in gently?

tmtmtmtm:
On the Site Manager we currently report if we’re actively fetching data for that instance from PopIt:

double flash 2015-04-10 at 13 05 59

However, all admin pages already report on this in the header bar.

So replace this message with the inprogress spinner:

inplace spinner 2015-04-12 at 10 59 44

(I suspect we should just be including that line of JS on all admin pages, rather than pulling it individually like this, but’s one for another change. #886)

Closes #839

### DIFF
--- a/nuntium/templates/nuntium/writeitinstance_update_form.html
+++ b/nuntium/templates/nuntium/writeitinstance_update_form.html
@@ -21,13 +21,13 @@ $(".chosen-person-select").chosen();
 {% block content %}
 
     <div class="page-header">
-        <h2>{% blocktrans with writeitinstance=writeitinstance %}Editing {{ writeitinstance }}{% endblocktrans %}</h2>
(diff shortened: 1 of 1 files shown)


chrismytton:
Looks good :+1:

PR #890, opened 2015-04-12
889 Tidy header bar text

### LINKED ISSUE #889: Replace "instance" and "WriteIt" in header-bar warnings
We still have a couple of occasions of "Instance" and "WriteIt" in the front-end: the warnings when there are no contacts, or when the site is has disallowed messages via the form. 

no contacts instance 2015-04-12 at 11 43 54

read only instance 2015-04-12 at 11 43 36

Replace these with "Site". It also doesn't seem to make sense to include the Test Mode banner in either of those cases, as no-one will be *able* to send messages anyway, so it doesn't matter who they would go to.

tmtmtmtm:
Replace “instance” and “WriteIt” with “Site”.

Also, only show the first message of: “Read Only”, “No Recipients”, “Test Mode” — there’s no point in showing “Test Mode” if either of the others are true.

read only 2015-04-12 at 11 50 13

no recipients 2015-04-12 at 11 50 03

test mode 2015-04-12 at 11 50 29

Closes #889

### DIFF
--- a/nuntium/templates/base_instance.html
+++ b/nuntium/templates/base_instance.html
@@ -94,8 +94,8 @@
   {% if writeitinstance and not writeitinstance.config.allow_messages_using_form %}
     <div class="site-alert">
         <div class="container">
-            <h3 class="site-alert__heading">{% trans "This instance is read-only." %}</h3>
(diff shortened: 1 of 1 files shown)


chrismytton:
Looks good :+1:

PR #892, opened 2015-04-12
887 When messages are disallowed

### LINKED ISSUE #887: Message Composer: Disable if allow_messages_using_form is unset
We currently give a message saying that the site is read only (which isn't strictly true, as it could be an API-driven site), but we still provide the "Send a Message" link the header bar, and the form still exists:

no form form 2015-04-12 at 11 26 45

We should also remove the "Write to this person" button on the Recipient page:

write to pedro 2015-04-12 at 12 12 32

tmtmtmtm:
**Remove “Send a message” from the header bar**

no messages link 2015-04-12 at 11 58 50

----

**Remove “Write to this person” from the Recipient page**

cant write to pedro 2015-04-12 at 12 12 44

----

**Disallow going directly to the message composer**

messages page disallowed 2015-04-12 at 12 10 24

Closes #887


chrismytton:
Looks good :+1:

PR #900, opened 2015-04-13
Ajax help balloons for management pages

zarino:
Builds on @davewhiteland’s prototype in the [help-pages-within-app](https://github.com/ciudadanointeligente/write-it/tree/help-pages-within-app) branch, but refactors his code to a) use Bootstrap for the popovers, b) use jQuery deferreds to speed up consecutive ajax hits to the same URL, and c) be a little bit simpler.

screen shot 2015-04-13 at 14 03 23

I've purposefully not styled the contents of the popover yet. I want to see what sort of stuff gets put into them, before I jump to any conclusions as to how they should be handled.

@davewhiteland – let me know if you're happy with the changes I've made, and if this still covers the use-cases you had in mind.


nuntium/templates/help/glossary.html
```diff
@@ -0,0 +1,71 @@
...
+
+  <dt>
+    <a name="testing-mode">testing mode</a>
+  </dt>
+  <dd id="help-testing-mode">
+    <p>
+      A WriteIt site in testing mode will <strong>not</strong> send messages
+      to the email addresses in its database &mdash; instead it will send
+      them to the site's owner. Also, WriteIt doesn't publicise testing sites:
+      whenever WriteIt dispolays a list of sites, testing sites are not shown
```
tmtmtmtm: typo: 'displays' (but we shouldn't hold up merging on fixing that if everything else is OK, as we'll be iterating quickly on this in the next couple of days anyway).

chrismytton:
Looks good :+1:

tmtmtmtm:
We need to remove all the "WriteIt" references here, but that can happen in a follow-on change.

PR #903, opened 2015-04-13
Issues/661 message template improvements

### LINKED ISSUE #661: Problems with email a Recipient gets
- [x] Subject line should just be the subject line 
- [x] Should use the instance name (and URL) instead of WriteIt
- [x] the explanation is backwards: instead of saying "you've received a message", it should say "X has sent you a message"
- [x] Then there's no need to include the Subject and From — we can just display the body
- [x] Should be say "you should reply" rather than "you can reply" :)
- [x] We should simply drop the "All the best, The WriteIt team" bit.
- [ ] URLs are broken
- [x] Should show the URL of the thread, not just of the site
- [x] Weird spacing and formatting throughout

---

Subject: `[WriteIT] Message: Test`
```
Dear Stubb Alexander

You’ve received a message via WriteIt, the online service that allows
anyone to write to politicians and other prominent people.


The message is as follows:
Subject: Test
From: Tony
Message body:
Test. Please ignore.


You can respond to Tony
simply by replying to this email.

** IMPORTANT ** Note that, as well as being sent to
Tony,
your response will be added to the
Eduskunta website at http://writeit.ciudadanointeligente.orghttp://eduskunta-2.writeit.ciudadanointeligente.org/en/,
where it will be public and visible to anyone.


If there’s a better email address for you, please let us know at
tony@mysociety.org,
where we’ll also be happy to answer any questions or problems.


All the best,
The WriteIt team
```

duncanparkes:
Improve the outbound message template to fix #661, and make it so that this is easier in future by not storing templates in the database unless they differ from the default.

Fixes #661

### DIFF
--- a/mailit/__init__.py
+++ b/mailit/__init__.py
@@ -1,4 +1,6 @@
+import itertools
 import logging
+import textwrap
 
 from django.conf import settings
 from django.core.mail import mail_admins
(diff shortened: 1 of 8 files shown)


chrismytton:
Looks good :+1:

tmtmtmtm:
I'm still worried by not knowing what the newly-skipped test line is, but merging for now.

Issue #904, opened 2015-04-13
Moderation: Move Accept/Reject button to the Message page, not the Message Manager


tmtmtmtm:
When deciding whether to Accept or Reject a message you need to look at the Message itself. So the buttons for whether to Accept or Reject it should be on the message page, not the Message Manager list view.

PR #905, opened 2015-04-13
Tidied this screen

wrightmartin:
## This is WIP, please do not merge 

- Messages now split into two tables, all messages, and messages in need of moderation (need dev help to do the logic here)
 - Removed moderation controls (they'll be on the message screen #904)
 - Public columns are now a single column with a checkbox toggle control (need a dev to wire this up)
 - Message count now includes a # to moderate (needs wiring up by a dev)

### DIFF
--- a/nuntium/static/sass/manager/_bootstrap.scss
+++ b/nuntium/static/sass/manager/_bootstrap.scss
@@ -15,7 +15,7 @@
 
 @import "../bootstrap/component-animations";
 @import "../bootstrap/dropdowns";
-// @import "../bootstrap/button-groups";
+@import "../bootstrap/button-groups";
 @import "../bootstrap/input-groups";
 @import "../bootstrap/navs";
 @import "../bootstrap/navbar";
--- a/nuntium/static/sass/manager/_style.scss
+++ b/nuntium/static/sass/manager/_style.scss
@@ -20,6 +20,13 @@ label {
     font-weight: normal;
 }
 
+caption {
+    border-bottom:1px solid $gray-lighter;
+    h3 {
+        font-size: 1.25em;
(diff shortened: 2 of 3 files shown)


wrightmartin:
screen shot 2015-04-13 at 16 27 31

tmtmtmtm:
@wrightmartin I tried to cherry pick the checkbox part of this out as a separate change, but I'm not sure how it's meant to work. In the current version, the eye is a `button`, so clicking it will actually submit the form, but does changing it to a checkbox mean we're going to need to actually add a submit button beside each one as well? (Each td is basically its own separate form at the minute, so we can't even just have a single submit button without changing how the whole thing works, AFAICS)

wrightmartin:
@tmtmtmtm hmm, no, I think it would work best as it was before, the checkbox should submit. @chrismytton @duncanparkes any ideas if this is possible?

zarino:
@tmtmtmtm @wrightmartin A `<button>` without a `type` attribute (as the page had before) will submit the form it's part of. If you want the same from a checkbox, you'll need to do it with javascript, and provide an actual submit button too as a fallback for non-javascript situations, like:

```html
<form>
  <input type="checkbox">
  <button type="submit">Save</button>
</form>
<script>
  $(function(){
    $(':checkbox').on('change', function(){
      $(this).parents('form').submit();
    }).siblings('button').hide();
  });
</script>
```

Or something.

tmtmtmtm:
I talked to Chris about this on the train this morning, and the best we could come up with that would actually still work plausibly well without JS was to have the default version be the button (as the current live version), and have the JS replace that with the checkbox, that it then also traps and auto-submits.

PR #906, opened 2015-04-13
Less makeshift email confirmation page

### LINKED ISSUE #733: "Confirmation Required" text and design
confirmation required 2015-04-07 at 12 33 02

We need better text for this stage.

zarino:
Fixes #733.

Includes some commented-out code showing how the sender’s email address could be passed to the template, and how we could provide a direct link to the most common email web clients. I think @chrismytton thought he might have a go at making that happen.

But, even in the case that we don't get the email correctly passed through, it degrades gracefully to this:

screen shot 2015-04-13 at 16 48 28


tmtmtmtm:
I've split out the template-only change as #921, which we should be able to push live quicker.

I'm also a little worried that for this approach we'll need to maintain a list of webmail providers, as different countries will presumably have local ones that lots of people use, and having to make code-level changes any time we want to add to that list doesn't seem ideal... (We don't need to solve that to put this version live, as this is definitely a Good Thing for lots of common cases; just musing aloud here in the hope that someone comes up with a good solution)

PR #920, opened 2015-04-13
Link to public message page where appropriate

### LINKED ISSUE #919: Message Manager: Add link to Public Message
At the minute you need to go to the admin page for the Message, and follow the link from there.

tmtmtmtm:
Add an extra column to the message manager that links to the public page if the message is public and confirmed.

thread link 2015-04-13 at 21 01 36

I suspect I also need to check something to do with moderation, but I don’t know how to do that, as message.moderated only seems to be true for messages that have *actively* been moderated, not ones that didn’t _need_ to be moderated. (And if we need that many checks, we should really just add a new 'visible?' method to the message. We also need to solve this problem on the Admin page for the message.)

I’m also not 100% convinced about adding this as a new column, but putting it beside the subject line doesn’t really work as then you have both the internal and external links side by side with no visual distinction between them. But @wrightmartin will be doing more work on this page tomorrow anyway, so at least this gives him something useful to move to elsewhere...

Closes #919

### DIFF
--- a/nuntium/templates/nuntium/profiles/messages_per_instance.html
+++ b/nuntium/templates/nuntium/profiles/messages_per_instance.html
@@ -35,7 +35,8 @@
         <thead>
           <tr>
             <th>{% trans "Message" %}</th>
-            <th class="text-center">{% trans "Replies"%}</th>
+            <th class="text-center">{% trans "Thread" %}</th>
+            <th class="text-center">{% trans "Replies" %}</th>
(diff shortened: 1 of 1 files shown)


chrismytton:
Looking at the moderation code I'm not entirely clear on the best way to check if a) moderation is enabled and if so b) checking if the message has been accepted.

So ignoring moderation, this looks fine :+1:

PR #921, opened 2015-04-13
733: Nicer "confirm your email" page

tmtmtmtm:
confirm your email 2015-04-13 at 21 16 28

Zarino has a better version of this change in #906, but it requires code-level changes, so let's just implement the template improvement first.


chrismytton:
Looks good :+1:

PR #928, opened 2015-04-14
Welcome page

### LINKED ISSUE #922: "Welcome To Your New Site" page
Creating a new site shouldn't just drop you straight into the Admin section for it, but should go to a Welcome page with a good introduction to what's going on.

(This will also have the side-effect of making it much more likely that we finish importing from your PopIt before you try to do anything else #908)

chrismytton:
This is the basic template for the welcome page. Needs populating with content but that'll get done in a separate pull request.

Part of #922

### DIFF
--- a/nuntium/templates/nuntium/profiles/welcome.html
+++ b/nuntium/templates/nuntium/profiles/welcome.html
@@ -1,10 +1,6 @@
 {% extends "base_manager.html" %}
 {% load i18n %}
 
-{% block header %}
-{% include 'nuntium/profiles/top_menu.html' with section='your-instances' %}
-{% endblock header %}
-
 {% block content %}
 
 {% if messages %}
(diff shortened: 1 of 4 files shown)


tmtmtmtm:
We should put this live as is, and iterate on the text. Anyone creating a site in the next hour or so will get a slightly odd page, but at this stage it's better to have it live IMO.

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -0,0 +1,24 @@
+{% extends "base_manager.html" %}
+{% load i18n %}
+
+{% block header %}
+{% include 'nuntium/profiles/top_menu.html' with section='your-instances' %}
```
duncanparkes: This is highlighting the wrong tab - I think you probably need to remove the `with section=`.

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -0,0 +1,24 @@
...
+{% endblock header %}
+
+{% block content %}
+
+{% if messages %}
```
duncanparkes: This messages stuff feels like something that should be in a base template. Perhaps move it to base_manager?

PR #940, opened 2015-04-14
922 welcome page text

davewhiteland:
New copy on the welcome page 

screen shot 2015-04-14 at 15 17 04

(Note to self: haven't included any config settings explicitly, could do so? I think they need passing in via the view if so.)

### DIFF
--- a/nuntium/templates/nuntium/profiles/welcome.html
+++ b/nuntium/templates/nuntium/profiles/welcome.html
@@ -55,7 +55,7 @@
   <li>
     <p>
       {% blocktrans %}
-        <a href="{{ url_template }}">update the email templates</a> so the
+        <a href="{{ url_template_update }}">update the email templates</a> so the
         messages your site sends out are exactly how you want them
       {% endblocktrans %}
     </p>
--- a/nuntium/templates/nuntium/profiles/welcome.html
+++ b/nuntium/templates/nuntium/profiles/welcome.html
@@ -16,9 +16,69 @@
 {% endif %}
 
 <div class="page-header">
-    <h2>{% trans "Welcome" %}</h2>
+    <h2>
+       {% blocktrans with name=writeitinstance.name %}
+         Welcome! This is {{ name }}
+       {% endblocktrans %}
+    </h2>
 </div>
 
-<p>{% trans "Write to your representative etc." %}</p>
+<p>
+  {% blocktrans with name=writeitinstance.name %}
+    This is your &ldquo;{{name}}&rdquo; message-sending site.
+  {% endblocktrans %}
+</p>
+<p>
+  {% blocktrans %}
+     Here are the things you need to check you've done before everything will work properly.
+  {% endblocktrans %}
+</p>
+<ul>
+  <li>
+    <p>
+      {% if writeitinstance.description %}
+        {% blocktrans with desc=writeitinstance.description %}
+          change the description (which is &ldquo;{{ desc }}&rdquo;)
+        {% endblocktrans %}
+      {% else %}
+        {% blocktrans with desc=writeitinstance.description %}
+          set the description
+        {% endblocktrans %}
+      {% endif %}
+    </p>
+  </li>
+  <li>
+    <p>
+      {% blocktrans %}
+        change the email templates so the messages your site sends out are
+        exactly how you want them
+      {% endblocktrans %}
+    </p>
+  </li>
+  <li>
+    <p>
+      {% blocktrans %}
+        pick the right options for things like:
+      {% endblocktrans %}
+    </p>
+    <ul>
+      <li>
+        <p>
+          {% blocktrans %}
+            how many recipients a message can go to (just one, or many?)
+          {% endblocktrans %}
+        </p>
+      </li>
+      <li>
+        <p>
+          {% blocktrans %}
+            how often to check for new data (which really means: do you
+            expect the data source to change?)
+          {% endblocktrans %}
+        </p>
+      </li>
+    </ul>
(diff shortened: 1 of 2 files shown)


davewhiteland:
PR providing text copy for #922

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -4,9 +4,107 @@
...
+      send real messages to real people!
+    {% endblocktrans %}
+  </p>
+{% endif %}
+<p>
+  {% if writeitinstance.persons.count %}
+    {% blocktrans count qty_pers=writeitinstance.persons.count %}
+      You've got only one recipient in your database.
+    {% plural %}
+      You've got {{ qty_pers }} recipients in your database.
```
tmtmtmtm: Might be nice to link this to the admin page of recipients?

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -4,9 +4,107 @@
...
+    {% endblocktrans %}
+  {% else %}
+    {% blocktrans %}
+      <strong>There aren't any recipients in your database</strong>
+      &mdash; you need to get some in there before anyone can send any messages!
+      <br>
+      Has your data source got recipients in it? Has it loaded OK?
```
tmtmtmtm: Surely we can tell them this? i.e. "Your data source still hasn't finished loading" vs "We loaded your data source, but couldn't find any recipients in it"?

davewhiteland:
I've made a bunch of local variables containing URLs here, which is inelegant: see
https://github.com/ciudadanointeligente/write-it/commit/ab6f08c7daf8210bd4f96e147f986bb9361fd799#diff-1ce295e72afef43dd0bfc386f14f5b7bR602

This is because we've overridden the `url` tag with one that has positional arguments, which prevents us using the {% url some-url as the_url %} syntax that offers `the_url` as a variable... this matters because I want to use the variable inside `blocktrans` tags (as the `href` target of an in-line link).
So by putting the URLs needed into the context as "flat" variables like this, they can appear in the `blocktrans` tags.

Horrible, sorry.

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -4,9 +4,107 @@
...
+    {% blocktrans %}
+      You're currently running in <strong>testing mode</strong>, so all
+      the emails sent will go to you (and not the recipients).
+      Contact us if you want this restriction lifted and you're ready to
+      send real messages to real people!
+    {% endblocktrans %}
+  </p>
+{% endif %}
+<p>
+  {% if writeitinstance.persons.count %}
```
tmtmtmtm: Ideally here we'd want to also check that there are *contactable* contacts, rather than just contacts (e.g. with no email addresses, or who aren't marked as Active)

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -4,9 +4,107 @@
...
+  {% blocktrans with name=writeitinstance.name %}
+    This is your &ldquo;{{name}}&rdquo; message-sending site.
+  {% endblocktrans %}
+</p>
+{% if writeitinstance.config.testing_mode %}
+  <p>
+    {% blocktrans %}
+      You're currently running in <strong>testing mode</strong>, so all
+      the emails sent will go to you (and not the recipients).
+      Contact us if you want this restriction lifted and you're ready to
```
tmtmtmtm: Let's be optimistic and assume this is a **when** rather than an **if** :)

nuntium/templates/nuntium/profiles/welcome.html
```diff
@@ -4,9 +4,107 @@
...
+            expect the 
+            <a href="{{ url_data_sources }}">data source</a> to change?)
+          {% endblocktrans %}
+        </p>
+      </li>
+      <li>
+        <p>
+          {% blocktrans %}
+            do you want to
+             <a href="{{ url_answernotification_update }}">receive a notification</a>
```
tmtmtmtm: Possibly worth being more explicit here with "receive a notification email"  (maybe even as strong as "a notification email yourself"?)

nuntium/user_section/views.py
```diff
@@ -596,3 +596,13 @@ class WelcomeView(DetailView):
     def dispatch(self, request, *args, **kwargs):
         self.kwargs['slug'] = request.subdomain
         return super(WelcomeView, self).dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super(WelcomeView, self).get_context_data(**kwargs)
+        # passing URLs in for easy insertion into the translation tags
```
tmtmtmtm: Might be worth noting that this is _required_ rather than just nice.

nuntium/templates/nuntium/profiles/welcome.html
davewhiteland: Yes, definitely, will add that now.

tmtmtmtm:
A few pretty small things, but the one that concerns me most is the "Contacts" text — if we're going to start doing different things based on the numbers of them, it should really be on contactable contacts, rather than just contacts as a whole.

tmtmtmtm:
We can check `persons_with_contacts` to see if we have contactable contacts rather than just contacts: I've just had to make the same change in https://github.com/ciudadanointeligente/write-it/commit/006f334f21474a37a6c475497345094c15bacdbb

tmtmtmtm:
I'll fix this up in another change. Putting this live to iterate from...

Issue #652, opened 2015-03-27
Provide an internal support email address


tmtmtmtm:
Contact the WriteInPublic team, for site administrators.

davewhiteland:
This is also an *inevitable* requirement (now) to have Testing Mode switched off, so this is critical now.

tmtmtmtm:
This can either be done as a direct mailto: or using a contact form (#944)

PR #941, opened 2015-04-14
remove link, add 'contact us' note

davewhiteland:
Do we have a useful target to make the "contact us" a link? Who is "us"?

...but as things stand, this is still better than having a link here :-)


tmtmtmtm:
Connected?: #652

tmtmtmtm:
I'm not 100% convinced that a dangling contact us with no explanation as to how is actually an improvement, but removing the link is good, and seeing the text all the time should hopefully prompt us to fix it, so :+1:

PR #947, opened 2015-04-14
Better explanation of how many people you can write to

### LINKED ISSUE #909: On the select a person to write to screen - search for up to five people text confusing
On this screen I wasn't sure if this meant that there were five people in the instance I could write to, or what this phrase even means - would it return up to five results, could I string together five names and search for them (and if the latter, is that something I would be likely to do) - I was confused

screenshot 2015-04-13 14 47 06

tmtmtmtm:
Explicitly say “You can send your message to up to {{ counter }} people” rather than combining it into “Search for up to {{counter }} recipients by name”.

multiple recipients 2015-04-14 at 16 55 56

Or, if you can only send to one person:

one recipient 2015-04-14 at 16 56 37

Fixes #909

### DIFF
--- a/nuntium/templates/write/who.html
+++ b/nuntium/templates/write/who.html
@@ -27,9 +27,10 @@
         <p class="form-group">
             <label for="id_who-persons">
                 {% blocktrans count counter=writeitinstance.config.maximum_recipients %}
-                    Search for recipient by name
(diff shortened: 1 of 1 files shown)


duncanparkes:
:+1:

Issue #933, opened 2015-04-14
Hide Moderation Options by default


tmtmtmtm:
Moderation isn't quite ready to go public (e.g. #923 #837 #704 #904) so let's keep it 'working' (FSVO), but hide all links to it. Then if someone really needs it we can show them how to use it (with suitable warnings) — but a normal admin won't try to use it.

tmtmtmtm:
Closed by #949

PR #949, opened 2015-04-14
Hide admin links to Moderation and Rate Limiting

### LINKED ISSUE #927: Hide rate-limiter admin page by default
Rate-limiting isn't quite ready to expose yet (#572 #416). Let's keep the admin page for it active, but not actually link it up anywhere. Then we can give the URL to anyone who actually asks for the functionality (and explain why it won't necessarily work well for them) without making it look it it's ready for use in the normal case.

tmtmtmtm:
We don’t want to link to Moderation or Rate Limiting for now as they’re not ready for public consumption.

The pages will still exist, so that we can tell people individually (with suitable warnings) that they can use them, but we don’t want to make them obvious to normal admins.

Before:

before 2015-04-14 at 17 19 59

After:

after 2015-04-14 at 17 19 29

Closes #927 and #933


duncanparkes:
:+1:

Issue #679, opened 2015-03-30
Different "web-based" vs "API-based" creation processes


tmtmtmtm:
Most instances will either be predominantly API-driven, or Web-driven — offer two different creation processes, each of which sets the relevant config options to sensible defaults for that approach.

chrismytton:
Need to add two extra settings, `WEB_BASED` and `API_BASED`. These can be set independently, but most often one will be set to `True` and the other to `False`. Setting both to the same boolean value will have undefined behaviour.

PR #951, opened 2015-04-14
Add settings for web vs API based sites

chrismytton:
This just adds the actual settings. We also need to actually use these
settings within the creation process etc so that they make the sites
behave in different ways.

Part of https://github.com/ciudadanointeligente/write-it/issues/679


tmtmtmtm:
Yay! :+1:

PR #952, opened 2015-04-14
751 Disable Instance

tmtmtmtm:
Move the “Allow Form” config option to be a “Disable Site” config on the Site Manager instead.


tmtmtmtm:
deleting in favour of #953

PR #953, opened 2015-04-14
751 Disable Instance

### LINKED ISSUE #751: Site Manager: Admins shouldn't be able to delete live sites

tmtmtmtm:
Move the “Web Based?” config option to be a “Disable Site” config on the Site Manager instead. 

Before:
site manager before 2015-04-14 at 21 47 38

After:
site manager after 2015-04-14 at 21 39 00

Then rename that form to make it more obvious what it's for (we still need better text for that, but we'll be doing all the admin text tomorrow):

Before:
allow form 2015-04-14 at 21 51 50

After:
site enabled 2015-04-14 at 21 47 07

This is the last "Advanced" option so we can now remove that section entirely:

Before:
about before 2015-04-14 at 21 47 49

After:
about after 2015-04-14 at 21 39 15

Closes #751

### DIFF
--- a/nuntium/templates/nuntium/profiles/manager-navigation.html
+++ b/nuntium/templates/nuntium/profiles/manager-navigation.html
@@ -3,10 +3,6 @@
 {% load staticfiles %}
   <ul class="navigation__list">
     <li class="{% if section == 'writeitinstance_basic_update' %}active{% endif %}"><a href="{% url 'writeitinstance_basic_update' subdomain=writeitinstance.slug %}">{% trans "About your site" %}</a></li>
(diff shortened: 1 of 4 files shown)


duncanparkes:
:+1:

PR #955, opened 2015-04-14
Remove link to Test Mode admin setting

### LINKED ISSUE #954: Remove link to Test Mode settings from public site
The public site still has a link to changing the Test Mode settings in the header bar if you're logged in as an admin.

test mode settings 2015-04-14 at 21 58 18

Those settings don't exist, so remove this link.

tmtmtmtm:
We no longer have an admin setting for Test Mode, so don’t link to it from the public site.

Before:
test mode settings 2015-04-14 at 21 58 18

After:

test mode settings after 2015-04-14 at 22 02 10

Closes #954


duncanparkes:
:+1:

Issue #530, opened 2015-02-24
Make sure "WriteIt" and "Instance" don't appear in the management backend


tmtmtmtm:
Users of a writeit instance care about that site — writing and reading messages, etc — not about the underlying software. Links to generic writeit things should be in the footer, and the header should be about the instance itself.

zarino:
This is a matter of making sure that the templates in question extend `base_instance.html`, and that the `writeitinstance` variable is passed in the template context.

Are there pages you have in mind, @tony, that currently display the "WriteIt" name in the top left corner, but which should display the instance name?

tmtmtmtm:
Once all current flurry of changes are merged in, I'll go through the whole site again in detail, but as a general rule of thumb, neither an an end-user (writing messages) nor an administrator, should *ever* see the word WriteIt anywhere (other than in a "How this site works" type setting). 

In the front-end the only that I can currently see is "A WriteIt instance" in the header (I think I'd just drop that completely in favour of a link in the footer to something that explains the Instance → WriteInPublic (etc) → WriteIt hierarchy.

There's also a "Your WriteIts" in the logged-in-user drop-down in the top right corner, but I'm classing that more as 'back-end', as a normal user will never see that (and will probably disappear anyway #604)

The word 'instance' should also ideally never leak out — and in the front end that's currently shown in the 'testing mode' banner. Again, under normal circumstances, a normal user should never see that, but that text will need changed anyway, as it's factually incorrect. (#706)

There are lots of places in the back-end interface, though, where these still need to change. Should we split out a separate ticket for that? (Or repurpose this ticket to be explicitly about the back-end?)

zarino:
I'm changing the title, since this ticket isn't just about the header any more.

It looks like most of the remaining references to "WriteIt" and "instance" are in the management pages.

tmtmtmtm:
#563 didn't give us anything better, so I think "Your Sites" has won for now for the places we really need a generic term.

tmtmtmtm:
We also need to make sure that the fallback of "WriteIt" for the top-level does the right thing when you're on WriteInPublic instead — e.g. in nuntium/templates/base_manager.html

PR #957, opened 2015-04-15
Change "Your WriteIts" to "Site Manager" in header bar

### LINKED ISSUE #956: Header bar still links to "Your WriteIts" if logged in
This should change from WriteIt to Site, per #530 — but it should also handle better the case of someone only having one site. However saying "Your Site" here implies that it will go to the front end, not the back end — so perhaps "Site Manager" would be better for all cases?

your writeits

tmtmtmtm:
removing “WriteIt” per #530, but also noting that you’re not going
directly to your Site(s), but to the Manager.

Before:

Back End:
back header before 2015-04-15 at 08 12 55

Front End:
front header before 2015-04-15 at 08 14 07

After:

Back End:
back header after 2015-04-15 at 08 12 33

Front End:
front header after 2015-04-15 at 08 05 36

Closes #956

### DIFF
--- a/nuntium/templates/base_instance.html
+++ b/nuntium/templates/base_instance.html
@@ -63,7 +63,7 @@
                             <li>
                                 <a href="{% url 'your-instances' subdomain=None %}">
                                     <span class="glyphicon glyphicon-tasks"></span>
-                                    {% trans "Your WriteIts" %}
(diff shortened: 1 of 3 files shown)


duncanparkes:
:+1:

PR #959, opened 2015-04-15
Recipient manager wording

tmtmtmtm:
The recipient manager still refers to “in your WriteIt”. Clean that up (for #530), remove an obsolete modal from when we could manually add contacts, and link to the Data Sources page when we mention it:

Before:
1 recipient before 2015-04-15 at 08 41 56
many recipients before 2015-04-15 at 08 41 51
0 recipients before 2015-04-15 at 08 41 46

After:
1 recipient after 2015-04-15 at 08 41 12
many recipients after 2015-04-15 at 08 41 06
0 recipients after 2015-04-15 at 08 32 34


duncanparkes:
:+1:

PR #960, opened 2015-04-15
Change default strapline

### LINKED ISSUE #741: Better default strapline?
Currently this is "Write to the people in power. Get answers. Be heard."

It can be customised on a per-instance basis, but we should consider whether there's a better default version.

tmtmtmtm:
Based on Myf’s text: “Write to the people who represent you. In public. Get answers. In public.”

new strapline 2015-04-15 at 08 55 29

Closes #741


duncanparkes:
Go on then! Some part of me doesn't like the <br/><br/>, but it works :+1:

tmtmtmtm:
Yeah, I'm not a huge fan of it either, but it'll do the job for now!

PR #965, opened 2015-04-15
Move moderation to the message admin

### LINKED ISSUE #904: Moderation: Move Accept/Reject button to the Message page, not the Message Manager

tmtmtmtm:
Currently moderation happens from the Messages page, rather than the
individual Message page. This is bad, as you want to be able to read
the message before moderating it. So move the button to there.

NB: This has all the same problems as the previous version, but but
it’s a slightly better version of broken here, and helps clean the
Messages Manager up a little bit more for other things we want to do
there.

Old:
needs moderated list old 2015-04-15 at 10 30 51
needs moderated message old 2015-04-15 at 10 30 59

New:
needs moderated 2015-04-15 at 10 21 07

accept it 2015-04-15 at 10 23 27

now moderated list 2015-04-15 at 10 23 37

now moderated message 2015-04-15 at 10 24 46

Closes #904

### DIFF
--- a/nuntium/templates/nuntium/profiles/message_detail.html
+++ b/nuntium/templates/nuntium/profiles/message_detail.html
@@ -27,14 +27,19 @@
           </dd>
 	  <dt>{% trans "Confirmed" %}</dt>
 	  <dd><i class="fa {% if message.confirmated %}fa-check{% else %}fa-times{% endif %}"></i></dd>
+
 	  {% if message.writeitinstance.config.moderation_needed_in_all_messages %}
 	  	<dt>{% trans "Moderated" %}</dt>
-        <dd><i class="fa {% if message.moderated %}fa-check{% else %}fa-times{% endif %}"></i></dd>
-      {% endif %}
-      {% if message.public %}
-	  <dt>{% trans "Public" %}</dt>
-	  <dd><i class="fa {% if message.public %}fa-check{% else %}fa-times{% endif %}"></i></dd>
-      {% endif %}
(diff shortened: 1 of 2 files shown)


nuntium/templates/nuntium/profiles/message_detail.html
```diff
@@ -27,14 +27,19 @@
...
+
 	  {% if message.writeitinstance.config.moderation_needed_in_all_messages %}
 	  	<dt>{% trans "Moderated" %}</dt>
-        <dd><i class="fa {% if message.moderated %}fa-check{% else %}fa-times{% endif %}"></i></dd>
-      {% endif %}
-      {% if message.public %}
-	  <dt>{% trans "Public" %}</dt>
-	  <dd><i class="fa {% if message.public %}fa-check{% else %}fa-times{% endif %}"></i></dd>
-      {% endif %}
+        {% if message.moderated %}
```
duncanparkes: Indentation seems a bit odd. Perhaps move these bits to the right a bit?

duncanparkes:
:+1:

nuntium/templates/nuntium/profiles/message_detail.html
tmtmtmtm: The indentation on this whole template is all over the place. We have a mixture of 2, 4, 6, and 8-space indentation at various places!

PR #971, opened 2015-04-15
968 no contactable contacts

### LINKED ISSUE #966: If site is read-only the "change this in settings" link is wrong
change read-only 2015-04-15 at 10 53 51

This goes to the basic admin, rather than the direct admin page

### LINKED ISSUE #968: No Recipients banner should show if you have no *contactable* recipients
http://empty.writeit.ciudadanointeligente.org/en/ is connected to http://wales.popit.mysociety.org/api/v0.1— this means I *do* have contacts: just none with email addresses.

This means I don't get the "No contacts" banner on the home page when I'm logged in, and will be left scratching my head as to why nothing is working:

empty 2015-04-15 at 11 07 22

tmtmtmtm:
If you have contacts in your instance, but none of them have email addresses, show the "No contacts" banner, as it'll get people closer to diagnosing the problem. (I still want to investigate what happens when you have _disabled_ all the contacts, but that's for another change #972 and much less likely to happen than someone connecting a PopIt with no emails). Closes #968 

Also, change the links in these banners to go directly to the relevant settings page, rather than the central admin page. Closes #966 

Before: (different URLs in bottom left corner when hovering on links)

read only before 2015-04-15 at 11 41 26
no recipients before 2015-04-15 at 11 41 13

After:
read only banner 2015-04-15 at 11 36 28

no recipients banner 2015-04-15 at 11 36 43

### DIFF
--- a/nuntium/templates/base_instance.html
+++ b/nuntium/templates/base_instance.html
@@ -86,7 +86,7 @@
             <h3 class="site-alert__heading">{% trans "This site is read-only." %}</h3>
             <p class="site-alert__description">{% trans "You can’t currently create new messages using this site." %}</p>
           {% if user.is_authenticated %}
-            <a class="site-alert__cta" href="{% url 'writeitinstance_basic_update' subdomain=writeitinstance.slug %}">{% trans "Change this in Settings" %}</a>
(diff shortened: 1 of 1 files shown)


tmtmtmtm:

PR #974, opened 2015-04-15
Show if a site is already disabled

### LINKED ISSUE #969: Sites Manager: Show whether site is enabled or disabled
We link to the page that lets you disable (or re-enable) the site, but don't say whether it's currently enabled or not.

tmtmtmtm:
In the instance manager, check if an instance is already Disabled before offering a link to disable it.

Before
disable before 2015-04-15 at 11 50 01

After
disable after 2015-04-15 at 12 00 54

It's a little ugly, but we haven't really solved the "show binary option with ability to change it" problem, so in the absence of design time, this'll have to do for now, especially as it should hopefully be a very rare occurrence.

Fixes #969

Also add new text from @MyfanwyNixon explaining the Test Mode section:

test mode explanation 2015-04-15 at 12 05 10

### DIFF
--- a/nuntium/templates/nuntium/profiles/your-instances.html
+++ b/nuntium/templates/nuntium/profiles/your-instances.html
@@ -70,7 +70,8 @@
       <hr />
       <div class="page-header">
         <h2>{% trans "Test Sites" %}</h2>
-        <p>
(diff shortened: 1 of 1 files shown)


tmtmtmtm:

PR #977, opened 2015-04-15
Pass ?next parameter through to python-social-auth

### LINKED ISSUE #963: New admin flow is confusing
From @MyfanwyNixon:

Signing in with Google takes me to my profile page, when I thought I was creating a new WriteIt site. Is it possible to land on http://writeit.ciudadanointeligente.org/en/writeitinstance/create instead?

chrismytton:
This means that if you arrive at the login page from another page, e.g. the create new instance page, then you will get redirected back to that page after signing up/in with Google.

Fixes #963


tmtmtmtm:
:+1:

PR #985, opened 2015-04-15
908 check import status

### LINKED ISSUE #908: Importing data from a PopIt instance, it appear to hang and not tell you when complete
I created a PopIt, imported the data (only two records) it did this, but didn't tell me and said that it was still importing - hanging.  By ignoring this and hitting "save changes" it appears to have worked, but as a user I would have sat waiting for ages

### LINKED ISSUE #979: Show count of Messages and People in the Admin bar
It would be nice to show the little indicators of how many people and messages you have in the sidebar, not just on the top-level instances list.

tmtmtmtm:
Make it obvious when a PopIt import finishes. 

If you’re on the Admin page for an instance, it shows a spinner if we’re currently fetching data for that instance. But, confusingly, nothing lets you know when that finishes, so you can sit for ages waiting for it to do something.

So have it only spin when the import is still in progress, and then also update the number of users in the sidebar.

(Which requires actually *adding* the number of users to the sidebar (closes #979); and whilst we’re there adding little icons beside each entry. Scope-creep-orama. The dependency management to do these all as separate changes at this point, especially with the afternoon Huboard delay) isn’t worth it.

about your site 0 2015-04-15 at 15 45 07

when finished, automatically becomes:

about your site 432 2015-04-15 at 15 51 52

Closes #908 . We might end up needing to do something similar on the Welcome To Your New site page, that has been added since this ticket was raised, but that depends on what that site is/does, as it doesn't really exist yet. #922

### DIFF
--- a/nuntium/templates/nuntium/profiles/manager-navigation.html
+++ b/nuntium/templates/nuntium/profiles/manager-navigation.html
@@ -2,11 +2,33 @@
 {% load subdomainurls %}
 {% load staticfiles %}
   <ul class="navigation__list">
-    <li class="{% if section == 'writeitinstance_basic_update' %}active{% endif %}"><a href="{% url 'writeitinstance_basic_update' subdomain=writeitinstance.slug %}">{% trans "About your site" %}</a></li>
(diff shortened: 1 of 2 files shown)


tmtmtmtm:

Issue #984, opened 2015-04-15
Source URL gets changed after site creation


tmtmtmtm:
If I follow the example given when creating a site:

![create a new site 2015-04-15 at 15 49 50](https://cloud.githubusercontent.com/assets/57483/7161450/18497c08-e387-11e4-975d-72b9ecdfb10f.png)

The URL that ends up in my data sources has switched from https to http, presumably at the same time as adding the /api part:

![date sources 2015-04-15 at 15 50 16](https://cloud.githubusercontent.com/assets/57483/7161459/257cd4ec-e387-11e4-87ed-9c43089c8e28.png)

Issue #981, opened 2015-04-15
Adding a new Popit doesn't resolve http vs https


tmtmtmtm:
Adding a new PopIt checks that you don't already have it linked, and Does The Right Thing depending on whether it's a base URL or API URL — *but* it doesn't check http vs https and lets you link both:

![http https 2015-04-15 at 14 44 17](https://cloud.githubusercontent.com/assets/57483/7160006/3e7c2672-e37e-11e4-862b-84159f90a901.png)

Issue #845, opened 2015-04-10
Site Manager: Default the WriteIt name to the PopIt name


tmtmtmtm:
When creating a new site, we should default the WriteIt name to be the same as the PopIt, and allow admins to change it, rather than making them pick both up front.

Issue #846, opened 2015-04-10
Store the name of a linked PopIt


tmtmtmtm:
We currently don't pay any attention to the name of a PopIt. We should fetch that too, so that we can

- [ ] display it on the Source manager (#817)
- [ ] default a new Site to have the same name as the PopIt (#845)

PR #988, opened 2015-04-15
Default the WriteIt name to the PopIt name

chrismytton:
This change is the first step towards https://github.com/ciudadanointeligente/write-it/issues/846 but in the interests of keeping it simple and making incremental changes this doesn't actually store the name of the popit anywhere, it just uses it as the default name for the site. If the popit doesn't have a name set (which is the default), then it uses the subdomain.

Fixes https://github.com/ciudadanointeligente/write-it/issues/845
Fixes https://github.com/ciudadanointeligente/write-it/issues/984
Fixes https://github.com/ciudadanointeligente/write-it/issues/995
Fixes https://github.com/ciudadanointeligente/write-it/issues/981


tmtmtmtm:
If the slug they pick already exists, this simply gives them a "eduskunta-2" etc (which they can't change later). 

This is still an improvement on what we have at the moment, so I'm OK with it going live (and #995 is fixed separately), but there's also a bunch of failing tests, so if it's easy to change this when fixing those, go for it!

tmtmtmtm:
Yay! :+1:

PR #1006, opened 2015-04-15
Add web_api_settings context processor

chrismytton:
This exposes `WEB_BASED` and `API_BASED` settings to the template so you
can do something like:

    {% if WEB_BASED %}
    Something
    {% endif %}

in the templates.


tmtmtmtm:
:+1:

PR #1028, opened 2015-04-16
Don't store default templates in the db.

### LINKED ISSUE #898: Don't store default email templates in the database
Currently, because we store the default email templates in the database it's tricky to tweak them, as we need to run migrations.

Instead we should only store the templates in the database if they've changed, and make the sending logic something like:

- Is there a template for this email in the database?
    - yes: use it
    - no: use the default template (from disk)

duncanparkes:
Stop storing default templates for content and subject in the
database for confirmation templates and new answer notification
templates.

The only place I know of where this still happens is the subject
of the actual message from mailit, and the default for this is
{subject}, so I don't think changing it is urgent!

Fixes #898

### DIFF
--- a/nuntium/migrations/0063_blank_new_answer_templates.py
+++ b/nuntium/migrations/0063_blank_new_answer_templates.py
@@ -1,0 +1,368 @@
+# -*- coding: utf-8 -*-
+import re
+
+from south.utils import datetime_utils as datetime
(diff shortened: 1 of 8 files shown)


tmtmtmtm:
Wahey!

chrismytton:
Looks good :+1:

PR #1030, opened 2015-04-16
Show the Admin's name rather than username in the header

### LINKED ISSUE #1024: Show admin's name, rather than username in headerbar
Registering via Google can create odd usernames (that the user has no control over). So instead of showing those in the header, use their name instead

your profile 2015-04-16 at 07 46 42

tmtmtmtm:
Before:
username 2015-04-16 at 08 26 52

After:
name 2015-04-16 at 08 24 12

Closes #1024


tmtmtmtm:
Wahey!

PR #1039, opened 2015-04-16
Check for existing slugs case-insensitively

### LINKED ISSUE #1012: slug checking should be case insensitive
When we check if a proposed slug is already taken we ignore case, so if 'eduskunta' is already taken, and I ask for 'eDuSkUnTa', it'll bypass the "that already exists", and create eduskunta-2 instead.

tmtmtmtm:
When we check if a proposed slug is already taken, make sure not to ignore case, so if 'eduskunta' is already taken, and someone asks for 'eDuSkUnTa', don’t let them proceed.

case insensitive screen shot 2015-04-16 at 17 49 40

Closes #1012


chrismytton:
Looks good :+1:

Issue #1046, opened 2015-04-17
Copy amends on your_instances


MyfanwyNixon:
1. On the "your_instances" page my copy reads "Use the menu on the left " but actually when you're on that page, you don't see the full menu. It needs to say:
 >  Click on 'edit' and then use the menu on the left.

2. "contact us to let us know" on the same page should be a link.

schlos:
Problem listed under point 2 is related to https://github.com/ciudadanointeligente/write-it/issues/1040

tmtmtmtm:
#652 is also done, so I think this can be closed.

PR #1047, opened 2015-04-18
Fixing 'Copy amends on your_instances' #1046

schlos:
This fixes ticket #1046 partially (only problem 1)


tmtmtmtm:
Thanks! There's something badly broken with Travis at the minute, which is making everything fail the build & test checks. Even though I don't see how this change could impact that in any way, I'm a little reluctant to force merge it. Hopefully the underlying problem gets resolved later today.

lfalvarez:
This looks great! :+1: I'm merging this as soon as travis says "Yes!"

lfalvarez:
In master already! =)

PR #1051, opened 2015-04-21
Clicking twice the confirmation url redirects to the message.

### LINKED ISSUE #961: Confirming a message twice gives an error
If a user clicks the confirmation link from email a second time, they get an error:

double confirmation 2015-04-15 at 09 30 24

Instead it should simply continue to displaying the thread as expected.

lfalvarez:
This instead of raising a 404 error.
Closes #961


tmtmtmtm:
@lfalvarez what's happening with this one?

lfalvarez:
@tmtmtmtm hey! thanks this was left behind and all the tests are passing and it should close #961

PR #1058, opened 2015-04-28
Fixing 'Copy amends on your_instances' #1046 (2)

schlos:
This fixes ticket #1046:
- problem nr.2 - contact us link is solved here
- problem nr.1 - text changes are fixed with PR #1047

### DIFF
--- a/nuntium/templates/nuntium/profiles/your-instances.html
+++ b/nuntium/templates/nuntium/profiles/your-instances.html
@@ -66,7 +66,7 @@
       <hr />
       <div class="page-header">
         <h2>{% trans "Test Sites" %}</h2>
-        <p>{% trans "Sites are set to Testing Mode by default. Use the 'Edit' link and make sure that your site is set up in the way that you want it. When you are ready, contact us to let us know and we’ll put your site live." %}</p>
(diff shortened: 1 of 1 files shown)


lfalvarez:
This looks awesome! I'm merging it as soon as I find out why all the tests are failing =)

lfalvarez:
Ok there is a slight problem here, the contact page is per instance so in this case for example you would have to provide a link for each instance rather than a general one. Any ideas how to do that?

I thought of something like this:

per instance

But of course I'm terrible using any kind of imaging software.

schlos:
Hi Luis,

Please check this out:
https://github.com/schlos/write-it/commit/d8af7b753058ceb68ef161d6f4a477f41ade1c4c

I've added this per instance, but I'm not sure if contact_page per instance is correct parameter.
Could someone check that on line 107? 
Just ping me if this is good and I'll make PR.

Cheers,
Miroslav

tmtmtmtm:
The solution here isn't to add multiple different contact links to the page — that would just über-confusing for users, especially as it's just going to a page that tells you the address to email, and which doesn't vary in any way on a per instance basis!

Instead we should simply fix it so that there's a non-instance-specific version of that page.

Or, if that is somehow too difficult (though I don't see why I would be) just have the header-bar link direct to the contact page on any one of their instances.

lfalvarez:
It isn't hard at all and it makes total sense to have a single page to ```contact us``` I think.

lfalvarez:
I'm closing this in favour of #1062

Issue #1070, opened 2015-05-22
Clicking on the create your own instance button on WIP or WI gives you a page not found error


paullenz:

tmtmtmtm:
Closed by #1071

PR #1071, opened 2015-05-22
Bug #1070 wrong create your own instance is broken.

lfalvarez:
What this does is set the default login url using the default language, which was causing but #1070.


tmtmtmtm:
is the CELERY_HAYSTACK_* stuff part of this fix, or has this slipped in from something unrelated?

lfalvarez:
@tmtmtmtm I think it wasn't part of the fix wait a sec

lfalvarez:
Taking that commit out raises a big bunch of problems. Anyhow I think it should go in a different PR.

PR #1079, opened 2015-07-02
Allow templates to be overridden on a subdomain-specific basis

davea:
This PR provides a middleware that inspects the subdomain for the current
request (as provided by django-subdomains) and inserts an entry for
subdomain-specific templates to settings.TEMPLATE_DIRS if one exists for
that subdomain.

I've not had to implement a custom template loader because `filesystem.Loader` seems to re-evaluate `TEMPLATE_DIRS` for each request, but that might not be the case in installations other than my local setup. It'd be good to have more eyes on this, specifically when `DEBUG = False`.

### DIFF
--- a/writeit/middleware.py
+++ b/writeit/middleware.py
@@ -1,0 +1,35 @@
+import os
+
+from django.conf import settings
+
+class SubdomainTemplateOverrideMiddleware(object):
+    """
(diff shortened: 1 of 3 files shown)


chrismytton:
Snap on ice :snowflake:

PR #1080, opened 2015-07-02
Set Google Analytics campaignName in base.html

chrismytton:
Make sure we set `campaignName` for subdomains in base.html as well as base_manager.html.


davea:
Looks good! :+1:

PR #1081, opened 2015-07-02
If theres a single contact skip first step

lfalvarez:
If there's a single person with contacts in an instance it automatically selects that person and redirects to the 'draft' step.


nuntium/views.py
```diff
@@ -85,6 +85,11 @@ def dispatch(self, request, *args, **kwargs):
 
     def get(self, *args, **kwargs):
         # Check that the form contains valid data
```
tmtmtmtm: You might want to move this comment down again, and add a comment about what this part is actually doing, but otherwise this all looks great!

tmtmtmtm:
A comment has drifted out of place in a slightly confusing way, but otherwise looks good! :+1:

Issue #862, opened 2015-04-11
Make it possible to create an instance from a third party site


tmtmtmtm:
e.g. EveryPolitician

chrismytton:
I started looking at this today, specifically to replace the PopIt url on the new site screen with a list of legislatures from EveryPolitician.

The main issue is that currently PopIt is quite deeply ingrained in the way things work. So to allow importing from other sources the `Contact` model first needs to be decoupled from the PopIt `Person` model. Other things that need tackling are the `WriteItInstanceCreateForm`, which currently inherits from `WriteItInstanceCreateFormPopitUrl`, which in turn calls `load_persons_from_a_popit_api` on the `WriteItInstance`. When people are being pulled from the PopIt API a `WriteitInstancePopitInstanceRecord` represents the current state of the import ("Not doing anything now", "Error", "Success", "Waiting and In progress"), so that will also need to be removed or replaced, depending on how long we think it might take to import from EveryPolitician (I suspect probably not long enough to warrant keeping track of it in such a granular way).

The main method in the `WriteItInstance` model that does the work of actually pulling people in from popit is `relate_with_persons_from_popit_api_instance`, which calls various methods in `nuntium/popit_api_instance.py` to pull the person objects from the remote popit API into the local database and then create some `Contact` objects with the local popit `Person` object related to it.

- [ ] Make the `person` attribute on the `Contact` model polymorphic, possibly using the [contenttypes framework](https://docs.djangoproject.com/en/1.6/ref/contrib/contenttypes/).
- [ ] On the "Create your new Site" screen replace the PopIt url with a dropdown list of legislatures from EveryPolitician's [`countries.json`](https://github.com/everypolitician/everypolitician-data/blob/master/countries.json)
- [ ] Use the EP term CSV to import a terms worth of people as `Contact` instances, associated with an object that looks like a `Person` model.

lfalvarez:
Hey! I'm sorry I'm reading this quite a few days after you created it.

So, here are my thoughts:
- I think it would be good to make some other forms of popolo available as well.
- So, what is the best approach to include all the countries as well as leave it as open for new sources of popolo formatted information?
- I really like the idea of importing CSV data from EP, would it make sense to use the ```django-popolo``` Person model ??

chrismytton:
[...]

PR #1102, opened 2015-10-22
Allow site creation from EveryPolitician data

chrismytton:
This allows you to create a site using data pulled from EveryPolitician. It currently does this by giving the user a dropdown list of countries, which are pulled from a proxy app running on heroku. This proxy app then provides a url which WriteInPublic can use to pull EveryPolitician data through an API which looks just like a PopIt API.

screen shot 2015-10-22 at 13 20 15

Part of #862

### DIFF
--- a/nuntium/templates/nuntium/create_new_writeitinstance.html
+++ b/nuntium/templates/nuntium/create_new_writeitinstance.html
@@ -21,10 +21,16 @@ politicians.</p>
       <option>Select a legislature</option>
       <option disabled></option>
       {% for country in countries %}
-      <option data-slug="{{ country.slug }}" data-email-count="{{ country.people_with_contact_details }}" value="{{ country.url }}">{{ country.country }} - {{ country.legislature }}</option>
+      <option
+      data-slug="{{ country.slug }}"
+      data-email-count="{{ country.people_with_contact_details }}"
+      data-person-count="{{ country.person_count }}"
+      value="{{ country.url }}"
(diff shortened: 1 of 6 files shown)


chrismytton:
@MyfanwyNixon We've updated the for for creating a new WriteInPublic site (see screenshot above). Could you possibly jazz up the copy a bit when you have a moment? Thanks!

nuntium/templates/nuntium/create_new_writeitinstance.html
```diff
@@ -13,38 +13,71 @@
...
+    var option = $('option:selected', this);
+    var slug = option.data('slug');
+    var emailCount = option.data('email-count');
+    var personCount = option.data('person-count');
+    if (emailCount > 0) {
+      submitButton.removeAttr('disabled');
+      $('#id_popit_url').val($(this).val());
+      $('#id_slug').val(slug);
+      if ((emailCount / personCount) <= 0.5) {
+        var text = 'We only have email addresses for ' + emailCount + ' out of ' + personCount + ' people.';
```
lfalvarez: My only concern is here where it seems that there is no translation for this javascript text. Anyhow this is not something for which I should not merge this into master.

lfalvarez:
I'm merging this !

tmtmtmtm:
Woohoo! :fireworks:

henare:
Very nice :ok_hand:

PR #1107, opened 2015-12-16
Go direct to drafting when given person ID in URL.

dracos:
This uses the popit/popolo ID as that will be known externally.


chrismytton:
Looks good to me! :+1:

PR #1109, opened 2015-12-18
Link direct to person

### LINKED ISSUE #1108: Link direct to person based on underlying Popolo ID
Currently the URLs for the people you can write to are in the form http://eduskunta.writeinpublic.com/en/per_person/20046/ — where the ID is the person's WriteIt ID. It would be good if there was a version that worked based on the ID of the underlying Popolo, so that these could be more easily constructed on external sites that share the same data source.

dracos:
Fixes #1108.


chrismytton:
Looks good! :+1:

PR #1112, opened 2015-12-24
Allow import of Popolo JSON file directly

### LINKED ISSUE #1105: Import from json instead of PopIt
It would be useful to have possibility to import data from Popolo-standardized json file instead of PopIt. For example, if the aim is to create an instance for a city council with a couple of representatives that are not going to change in the next 4 years, setting up a new instance of PopiIt seems to me as an overkill (if all we need is to import a couple of dozens of strings).

### LINKED ISSUE #1106: Adding new data sources shouldn't refer to PopIt
`/manage/settings/sources/` offers a box to link to a new PopIt:

screen shot 2015-12-16 at 14 57 36

This should, like the front page, be a choice between EP legislatures, or the URL to your own Popolo.

### LINKED ISSUE #1111: Country dropdown fails without JavaScript
The dropdown is shown, but picking from it and submitting never succeeds.
The server should accept an entry in the dropdown and use that equivalent to an entry in the text box.

dracos:
It now accepts as well as a popit-style URL (to which it would add /persons and expect paginated data), a popolo JSON file with direct data in persons. The popolo JSON for people should contain id, name, (email), (image), (summary) and memberships either inline or in separate memberships.

Fix some JS-only/JS errors along the way.

Fixes #1105. Fixes #1111.  Fixes #1106. Might be enough for #862.

~~Test failure is in downloading Django, don't know what that's about; passes locally.~~

### DIFF
--- a/nuntium/forms.py
+++ b/nuntium/forms.py
@@ -162,7 +162,7 @@ class PopitParsingFormMixin(object):
 class WriteItInstanceCreateFormPopitUrl(ModelForm, PopitParsingFormMixin):
     popit_url = URLField(
         label=_('Popolo URL'),
-        help_text=_("Example: https://eduskunta.popit.mysociety.org/"),
+        help_text=_("Example: https://cdn.rawgit.com/everypolitician/everypolitician-data/1460373/data/Abkhazia/Assembly/ep-popolo-v1.0.json"),
         required=True,
         )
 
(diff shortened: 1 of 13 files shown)


chrismytton:
Looks good to me :+1:

michalskop:
Is there a documentation/example how to use it?

dracos:
@michalskop On the writeinpublic create page, you can select "I have my own suitably formatted popolo" and supply the URL of a Popolo JSON file, and it gives you an example file there. It should hopefully be just what you asked for in #1105 :) 

In terms of what parts of a Popolo JSON file it currently looks at, it looks at the top-level "persons" key, and either inline memberships on each person, or a separate top-level "memberships" key. It uses the id, name, email (or contact details email entry), image, and summary attributes of a person.

michalskop:
@dracos worked fine for me here:  http://plasy-test.writeinpublic.com/
Much simpler now.

PR #1113, opened 2016-01-08
Czech localization file (based on Hungarian one)

michalskop:


chrismytton:
Couple of screenshots of this in action:

screen shot 2016-01-08 at 11 45 02

screen shot 2016-01-08 at 11 45 11

chrismytton:
@michalskop This is great! Thanks for taking the time to do this :+1:

chrismytton:
This is now live http://writeinpublic.com/cs/

Issue #421, opened 2015-01-13
Document webhooks


tmtmtmtm:

henare:
These were added in #191.

henare:
The documentation added in #1125 is sufficient so I think this issue now needs to be about exposing the settings page in the menu.

tmtmtmtm:
I think that as a first pass, these should only appear in the menu for sites that are API_BASED rather than WEB_BASED, grouped with the other API docs — they'll likely just be too confusing for people simply running a hosted WriteInPublic.

equivalentideas:
I'm glad this is an issue already, I just went looking and couldn't for the life of me find the webhooks settings page 🔍 ¯\_(ツ)_/¯

Issue #414, opened 2015-01-13
API docs


tmtmtmtm:
* shouldn't just be in a [gist](https://gist.github.com/lfalvarez/b9506f54037e9ceaa823)
* Why do calls return both 'persons' and 'people'? And why are those different to each other?
* The popit_url seems wrong (no instance; and doesn't match the popit_id)

lfalvarez:
hey @henare this is the issue mentioned in the mails =)

lfalvarez:
Sorry it took me forever to realize that there already was a place where this sort of documentation is placed (very hidden due to the quality of it) but if you go to http://<your-instance-slug>.127.0.0.1.xip.io:8000/en/manage/settings/api/ you'll find a bit of it.
Anyhow, I'm trying to improve a bit this docs.

henare:
Thanks Felipe! Is there any documentation of webhooks to notify you of a
new reply?

henare:
>Is there any documentation of webhooks to notify you of a new reply?

There's another issue for this, see #683.

lfalvarez:
Uhmm sorry @henare there is very little done about webhooks. They work and everything but there's no maintainer (apart from the one provided by Django admin).

PR #1125, opened 2016-01-15
Webhooks maintainer

lfalvarez:
This adds a place in the backend for creating and listing webhooks.
Related  to #414 and #421.


henare:
I just tested this on writeit.ciudadanointeligente.org - very nice :sparkles: 

I'm sure you know but it's not shown on the sidebar menu - you need to manually type the address.

lfalvarez:
Yeah! Is just that I didn't know where to put it in the sidebar I think we could open another ticket proposing where in the sidebar, the webhooks and the documentation about the api should go.

PR #1130, opened 2016-02-18
Integration guide

henare:
This is an integration guide. As mentioned in this thread: https://groups.google.com/forum/#!topic/poplus/JFwRZ4ENj1k

TODO:
- [x] How to disable new answer notifications
- [x] Creating and using webhooks to receive messages
- [x] Link to it from somewhere (the readme?)

### DIFF
--- a/INTEGRATION_GUIDE.md
+++ b/INTEGRATION_GUIDE.md
@@ -14,7 +14,7 @@ https://australia-representatives.writeit.ciudadanointeligente.org/manage/settin
 
 You should see a screen like this:
 
-TODO: Screenshot
+![API screenshot](http://i.imgur.com/tEHefQu.png)
 
 It contains all the details you need for sending a message via the API. You can use a `curl` command similar to the following to test this from the command line:
 
--- a/INTEGRATION_GUIDE.md
+++ b/INTEGRATION_GUIDE.md
@@ -49,7 +49,7 @@ https://australia-representatives.writeit.ciudadanointeligente.org/manage/api/se
 
 If WriteIt is integrated into your application you probably want your application to be responsible for sending new answer notifications instead of them being sent by WriteIt. When your application receives a message from WriteIt it should notify the original sender that they have a new message and to visit your application to see it.
 
-To achieve this set the `author_email` to an administrator mailbox you control when you create new messgaes via the API. This will redirect these WriteIt new answer messages to you and is a handy check that your application should be getting messages from WriteIt.
+To achieve this set the `author_email` to an administrator mailbox you control when you create new messages via the API. This will redirect these WriteIt new answer messages to you and is a handy check that your application should be getting messages from WriteIt.
 
 ## Receiving messages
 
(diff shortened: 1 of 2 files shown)


henare:
This is no longer WIP and is ready for review. It's based on our experience of setting up WriteIt with PlanningAlerts, which is now live.

henare:
The Travis build has failed but as I've only changed Markdown files and `master` is passing this failure must be a test isolation failure. So rather than restarting the build I'll leave it as broken so someone else can extract that into an issue and merge this :smile:

INTEGRATION_GUIDE.md
```diff
@@ -0,0 +1,70 @@
...
+
+This template should probably be emptied out to only contain:
+
+    {content}
+
+This will give your application the greatest control over the messages it sends as you will be able to specify the full content and subject in your API call.
+
+### Confirmation Template
+
+WriteIt should be configured by default to not require confirmation for messages sent via the API, therefore this template will not be used. You can check this setting at the `/manage/api/settings/` address for your site, e.g.
```
tmtmtmtm: I stumbled over the "should" here, as that sort of wording is usually in the context of telling me something that _I_ should do. This might be a little clearer as something like "By default WriteIt is configured to not require…"

INTEGRATION_GUIDE.md
```diff
@@ -0,0 +1,70 @@
...
+
+https://australia-representatives.writeit.ciudadanointeligente.org/manage/settings/webhooks/
+
+Details of what is sent in the payload of the webhook is shown on that page.
+
+### Fetching messages and all their replies
+
+If the webhook payload does not contain all the reply details you need you can use the webhook as a way to trigger fetching the full message and replies from the API.
+
+This is how [PlanningAlerts](https://github.com/openaustralia/planningalerts/) integrates with WriteIt. When the application receives a webhook it uses the `message_id` from the payload to do a full fetch of the message from the API.
```
tmtmtmtm: Might be nice to link directly to the code that's doing this, so that people can see by example how this works.

INTEGRATION_GUIDE.md
henare: That makes it clearer :+1:

INTEGRATION_GUIDE.md
henare: I'm not sure if it helps right now because [the code](https://github.com/openaustralia/planningalerts/blob/master/app/models/comment.rb#L62-L82) is pretty messy and very application-specific. What do you think?

tmtmtmtm: I think a majority of people learn better by example, and this example doesn't look particularly bad to me… but maybe @chrismytton or @lfalvarez might have a different opinion?

lfalvarez: I think it's actually pretty good and it's quite clear as well :+1:

henare: Thanks @tmtmtmtm @lfalvarez, I've added links to the examples and I agree it's better now.

tmtmtmtm:
@henare thanks a bunch for pulling this together, and apologies that it's been ignored for forever. Whatever issue travis had with it in the past seems to have been transient, so I think this is good to go.

PR #1135, opened 2016-02-23
Update vagrant JDK install to match elasticsearch expected version

struan:
The current version of elasticsearch relies on Java 7 and silently fails
to launch which Java 6 so install Open JDK 7


chrismytton:
:+1:

PR #1136, opened 2016-02-23
Put tooltip activation javascript in base admin template

### LINKED ISSUE #886: Include the data-tooltip JS on all admin pages
Currently any template that wants to use tooltips needs to include

```
<script type="text/javascript">
  $(function () {
    $('[data-toggle="tooltip"]').tooltip()
  })
</script>
```

Instead, we should simply add this to all admin pages.

struan:
Avoids duplicating the code across individual templates

Fixes #886

### DIFF
--- a/nuntium/templates/base_manager.html
+++ b/nuntium/templates/base_manager.html
@@ -127,6 +127,9 @@
               }
           }
       });
+      $(function () {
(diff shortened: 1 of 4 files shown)


chrismytton:
Looks good :+1:

PR #1137, opened 2016-02-23
Add can_create_messages property to instances

### LINKED ISSUE #891: Add standard method for can_create_messages
Currently we have to check `if writeitinstance.contacts.exists and writeitinstance.config.allow_messages_using_form` in multiple places.

Provide a single method that encapsulates this.

struan:
Use this in the templates to determine if we can send messages rather
than checking for contacts and enabled.

Fixes #891


chrismytton:
Looks good :+1:

PR #1138, opened 2016-02-23
Allow user to swap back to legislature list on instance create page

### LINKED ISSUE #1126: add ability to redraw legislature list
when you are creating a new instance, 
first show a list of country and all seem ok, but if you click on the link called "I have my own suitably formatted popolo" the list change to an input field and you can't go back for the country list.

struan:
Add a link to re-show the legislature list and hide the manual popolo
list.

Fixes #1126


chrismytton:
Awesome! :+1:

Issue #1091, opened 2015-07-03
Set a default language based on a config in the instance


lfalvarez:
So for example in http://infoszab.writeinpublic.com would try to automatically set Hungarian as the default language.

tmtmtmtm:
Suggested steps (via @dracos)

* Add a database column (presumably to WriteItInstanceConfig model) to store the language code for that instance
* Subclass Django's LocaleMiddleware and add a step to its process_request function to, if on an instance subdomain, add a fallback to that instance's language rather than just the LANG_CODE.
* Use that middleware rather than the built-in one.

tmtmtmtm:
@michalskop do you already have some instances already created that you'd like us to test this with?

michalskop:
http://plasy-2014.writeinpublic.com/

PR #1141, opened 2016-03-30
Allow instances to set a default language

chrismytton:
screen shot 2016-03-30 at 16 18 30

This adds a piece of middleware which checks the instance for a default language and sets that on the request if there isn't a translation for the users' default language.

Fixes https://github.com/ciudadanointeligente/write-it/issues/1091

### DIFF
--- a/nuntium/middleware.py
+++ b/nuntium/middleware.py
@@ -13,9 +13,21 @@ class InstanceLocaleMiddleware(LocaleMiddleware):
         except WriteItInstanceConfig.DoesNotExist:
             default_language = settings.LANGUAGE_CODE
 
+        # Django 1.6 doesn't provide a nice way to hook into the language
+        # selection process to override it with a default. So instead we add
+        # a default language to the session and then remove it again after
+        # running 'get_language_from_request'.
+        did_override_session_language = False
+
         if not request.session.get('django_language', None):
             request.session['django_language'] = default_language
+            did_override_session_language = True
+
         language = translation.get_language_from_request(
             request, check_path=check_path)
+
+        if did_override_session_language:
+            del request.session['django_language']
+
         translation.activate(language)
         request.LANGUAGE_CODE = translation.get_language()
--- a/nuntium/middleware.py
+++ b/nuntium/middleware.py
@@ -1,0 +1,21 @@
+from django.middleware.locale import LocaleMiddleware
+from django.utils import translation
+from django.conf import settings
+
+from nuntium.models import WriteItInstanceConfig
+
+
+class InstanceLocaleMiddleware(LocaleMiddleware):
+    def process_request(self, request):
+        check_path = self.is_language_prefix_patterns_used()
+        try:
+            default_language = WriteItInstanceConfig.objects.get(writeitinstance__slug=request.subdomain).default_language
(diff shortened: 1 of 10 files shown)


nuntium/middleware.py
```diff
@@ -0,0 +1,33 @@
...
+        try:
+            default_language = WriteItInstanceConfig.objects.get(writeitinstance__slug=request.subdomain).default_language
+        except WriteItInstanceConfig.DoesNotExist:
+            default_language = settings.LANGUAGE_CODE
+
+        # Django 1.6 doesn't provide a nice way to hook into the language
+        # selection process to override it with a default. So instead we add
+        # a default language to the session and then remove it again after
+        # running 'get_language_from_request'.
+        did_override_session_language = False
```
dracos: I'm assuming you want this "default" to take precedence to a user's browser setting, but still overrideable by the URL; in which case, this is okay, though I wonder if it might be clearer to copy the first bits of `get_language_from_request` into here? Then if there's a language in the URL it doesn't have to do any database look up. Something like:

```python
def process_request(self, request):
   """Same as parent, except calling our own get_language_from_request"""
    check_path = self.is_language_prefix_patterns_used()
    language = self.get_language_from_request(
        request, check_path=check_path)
    translation.activate(language)
    request.LANGUAGE_CODE = translation.get_language()

def get_language_from_request(request, check_path=False):
    from django.conf import settings
    supported = SortedDict(settings.LANGUAGES)

    if check_path:
        lang_code = translation.get_language_from_path(request.path_info, supported)
        if lang_code is not None:
            return lang_code

    try:
        lang_code = WriteItInstanceConfig.objects.get(writeitinstance__slug=request.subdomain).default_language
    except WriteItInstanceConfig.DoesNotExist:
        lang_code = None

    # Same behaviour as session entry – (Django aside: not sure why this doesn't use get_supported_language_variant like cookie does?)
    if lang_code in supported and lang_code is not None and translation.check_for_language(lang_code):
        return lang_code

    # Call with check_path False as we've already done that above!
    return translation.get_language_from_request(request, check_path=False)
```

nuntium/middleware.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if not request.session.get('django_language', None):
```
dracos: This will change if Django is updated (it's now `_language`), so I'd add a note to this effect.

nuntium/user_section/forms.py
```diff
@@ -41,6 +42,18 @@
 
 
 class WriteItInstanceBasicForm(ModelForm):
+    language = ChoiceField(choices=settings.LANGUAGES)
+
+    def __init__(self, *args, **kwargs):
+        kwargs['initial']['language'] = kwargs['instance'].config.default_language
```
dracos: If this form is ever used for creation, not just editing, instance won't exist. Could just check it exists before setting this line?

dracos:
Current test issues (ah, I see you've committed something for that already) and new tests missing as you say aside, looks okay to me :)

chrismytton:
@dracos Thanks for the review. We went with your suggestion in the end and overrode `get_language_from_request`, much nicer than messing around with the session! Have also addressed your other points and added some tests.

Issue #718, opened 2015-04-03
Update to Django 1.8


duncanparkes:
Now that Django 1.8 has been released, 1.6 is no longer supported, and won't get security fixes. We should update to 1.8, which is the new LTS release.

lfalvarez:
What does it take to upgrade to Django 1.8? change something with migrations?

duncanparkes:
I usually go through the changelog with a fine-toothed comb grepping the codebase for anything which I think might be a problem. Having said that, you have a pretty good test suite, so perhaps we trust that it would work sufficiently well if all the tests pass?

PR #1158, opened 2016-07-05
Pin django-nose to the last version that works with Django < 1.7

mhl:
As of django-nose/django-nose@7f205f1b16e641a7b1 support for Django
earlier than 1.7 was dropped, but write-it is still using Django
1.6.10 (which is no longer supported, but that's another issue: #718)
django-nose 1.4.3 is the last version that works with Django 1.6.


chrismytton:
Looks good 👍

PR #1160, opened 2016-07-05
Switch to using container-based builds on Travis

mhl:
These are much faster to start up than builds on the legacy
infrastructure; the .travis.yml doesn't use sudo anyway, so this should
work fine.


chrismytton:
🚀 👍

PR #1162, opened 2016-07-06
Move some models to a new instance app to avoid circular dependencies

mhl:
There was an unnecessary circular dependency between the 'nuntium' and
'contactos' Django apps:
    
   contactos.Contact needs nuntium (for nuntium.WriteItInstance)
   nuntium.OutboundMessage needs contactos (for contactos.Contact)
    
This wouldn't be a big deal, except that when we're migrating to Django
1.7 it means that makemigrations generates multiple initial migrations
(0001_initial and 0002_auto..., the latter of which adds the problematic
foreign keys). The second of these migrations didn't work properly on
Django 1.7 and SQLite, so the easiest way to deal with that upgrade
seems to be to:
    
 * Remove the circular dependencies with South migrations and
   Django 1.6
 * Then do the upgrade to Django 1.7, when we should now be able
   to have single initial migrations
    
This pull request deals with the first of those by creating a new
'instance' app, which contains:
    
  WriteItInstance
  Membership
  WriteItInstanceConfig
  WriteitInstancePopitInstanceRecord

### DIFF
--- a/contactos/migrations/0008_move_to_instance.py
+++ b/contactos/migrations/0008_move_to_instance.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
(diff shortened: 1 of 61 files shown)


chrismytton:
Tried this with sqlite and mysql and it's working as expected. So :+1: from me.

PR #1163, opened 2016-07-06
Upgrade to Django 1.7.11 (depends on #1162)

mhl:
Once #1162 has been merged, deployed to all of our installations
of write-it, and we're happy that's working fine, this pull request
should do the upgrade to Django 1.7.

This is a fairly significant upgrade, since it's the one where we switch
from database migrations based on South to its replacement, the new
Django core migrations. It's best to do this upgrade and check it's all
working and then do the upgrade to Django 1.8 afterwards.

The trickiest bit of this is that it requires an update to django-tastypie
which needs some corresponding updates to write-it's tastypie resources.

Note that I haven't been exhaustively through
everything in the release notes;
since there's such good test coverage, hopefully this has caught any
major problems.

Connects to: #718


chrismytton:
All looks to be in working order 👍  🚀

PR #1164, opened 2016-07-06
Allow anonymous users

struan:
Changes required to allow users to submit messages with no author name attached.

### DIFF
--- a/nuntium/forms.py
+++ b/nuntium/forms.py
@@ -108,9 +108,18 @@ class DraftForm(ModelForm):
         widgets = {
             'content': Textarea(attrs={'class': 'form-control', 'required': 'True'}),
             'subject': TextInput(attrs={'class': 'form-control', 'required': 'True'}),
-            'author_name': TextInput(attrs={'class': 'form-control', 'required': 'True'}),
+            'author_name': TextInput(attrs={'class': 'form-control'}),
             'author_email': EmailInput(attrs={'class': 'form-control', 'required': 'True'}),
         }
+
+    def __init__(self, allow_anonymous_messages, *args, **kwargs):
+        super(DraftForm, self).__init__(*args, **kwargs)
+        self.allow_anonymous_messages = allow_anonymous_messages
+
+    def clean_author_name(self):
+        if not self.allow_anonymous_messages and self.cleaned_data['author_name'] == '':
+            raise ValidationError(_('Author name is required'), code='name_required')
+        return self.cleaned_data['author_name']
 
 
 class PreviewForm(ModelForm):
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -108,7 +108,7 @@ class Message(models.Model):
     """Message: Class that contain the info for a model, \
     despite of the input and the output channels. Subject \
     and content are mandatory fields"""
-    author_name = models.CharField(max_length=512)
+    author_name = models.CharField(max_length=512, default='', blank=True)
     author_email = models.EmailField()
     subject = models.CharField(max_length=255)
     content = models.TextField()
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -176,6 +176,13 @@ class Message(models.Model):
         return people
 
     @property
+    def author_name_for_display(self):
+        name = self.author_name
+        if name == '':
+            name = _('Anonymous')
+        return name
+
+    @property
     def outbound_messages(self):
         no_contact_oms = NoContactOM.objects.filter(message=self)
         outbound_messages = OutboundMessage.objects.filter(message=self)
@@ -264,7 +271,7 @@ class Message(models.Model):
 
         context = {
             'owner_name': self.writeitinstance.owner.username,
-            'author_name': self.author_name,
+            'author_name': self.author_name_for_display,
             'author_email': self.author_email,
             'recipients': u', '.join([x.name for x in self.people]),
             'subject': self.subject,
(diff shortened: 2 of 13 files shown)


struan:
At the moment this doesn't link from Anonymous messages to other messages by the same user, and indeed, if you try to do so you get a 404. We need to decide if that's the behaviour we want.

Also, I suggest waiting till the Django 1.7 upgrade work is complete and then we can update the migration in this before merging.

nuntium/models.py
```diff
@@ -108,7 +108,7 @@ class Message(models.Model):
     """Message: Class that contain the info for a model, \
     despite of the input and the output channels. Subject \
     and content are mandatory fields"""
-    author_name = models.CharField(max_length=512)
+    author_name = models.CharField(max_length=512, default='', blank=True)
```
mhl: I think Django will want to generate a migration for this change, but it's not in this PR.

nuntium/forms.py
```diff
@@ -108,10 +108,19 @@ class Meta:
...
+            'author_name': TextInput(attrs={'class': 'form-control'}),
             'author_email': EmailInput(attrs={'class': 'form-control', 'required': 'True'}),
         }
 
+    def __init__(self, allow_anonymous_messages, *args, **kwargs):
+        super(DraftForm, self).__init__(*args, **kwargs)
+        self.allow_anonymous_messages = allow_anonymous_messages
+
+    def clean_author_name(self):
+        if not self.allow_anonymous_messages and self.cleaned_data['author_name'] == '':
```
mhl: Maybe this is clearer as:

```python
    if not (self.allow_anonymous_messages or self.cleaned_data['author_name']):
```

... or that might just be me.

nuntium/tests/see_all_messages_from_a_person_tests.py
```diff
@@ -45,6 +45,46 @@ def test_get_all_messages_from_a_person_view(self):
...
+        url = reverse('all-messages-from-the-same-author-as', subdomain=self.message.writeitinstance.slug,
+            kwargs={
+                'message_slug': self.message.slug
+            })
+
+        response = self.client.get(url)
+        self.assertEquals(response.status_code, 200)
+        self.assertIn('message_list', response.context)
+        self.assertEquals(len(response.context['message_list']), 1)
+        self.assertEquals(response.context['message_list'][0].id, 2)
```
mhl: As well as just checking the length of the messages here, for safety I'd add an `assertNotIn('A message', response)` (and similarly for the next test) to make sure that's not listed even if someone changes the test setup.

nuntium/views.py
```diff
@@ -264,6 +264,10 @@ def get_queryset(self):
         qs = qs.filter(writeitinstance=self.writeitinstance).\
             filter(public=True).filter(confirmated=True).\
             filter(author_email=self.message.author_email)
+        if self.message.author_name == '':
+            qs = qs.filter(author_name='')
+        else:
+            qs = qs.exclude(author_name='')
```
mhl: I would be more comfortable with not allowing messages which are anonymous from being linked to others from the same email address via this view. I can imagine that the clustering of interests of a correspondent might reduce their anonymity. Maybe there's already been some discussion of this though?

struan: Isn't that what 	29a34b6 does?

nuntium/tests/see_all_messages_from_a_person_tests.py
```diff
@@ -65,6 +66,7 @@ def test_anonymous_messages_not_listed_from_non_anonymous_link(self):
         self.assertEquals(len(response.context['message_list']), 1)
         self.assertEquals(response.context['message_list'][0].id, 2)
 
+    @skip("we don't allow links to all authors for anonymous messages")
```
mhl: Sorry, I don't follow why this test is being skipped now?

mhl: Oh, because the links would have been on a page that you're making 404

mhl:
What's here looks OK to me - I've just left a few small comments. However, by its nature it's a tough one to be sure that there's nothing missed that might affect anonymization. (I mean, I've grepped for `author_name` and didn't spot anything else significant, but could easily have missed things that way.)

nuntium/views.py
mhl: Yep, but I was reviewing commit-by-commit so hadn't got there yet :)

JenMysoc:
Can this be merged now?

mhl:
This was merged a while ago - the branch must have been rewritten and merged without updating this topic branch. The merged commits are:

* 3ffb190 - add some extra checks to anonymous thread tests (2 weeks ago) \<Struan Donald>
* 983e187 - test for notification of answer to anonymous message (2 weeks ago) \<Struan Donald>
* a89c23d - add and use author_name_for_display property (2 weeks ago) \<Struan Donald>
* 9ac71e6 - handle anonymous users in message detail title (2 weeks ago) \<Struan Donald>
* e4109a7 - return a 404 for all messages by person for anonymous messages (2 weeks ago) \<Struan Donald>
* 050f5a7 - test From on message thread correct for anonymous messages (2 weeks ago) \<Struan Donald>
* 3e63032 - do not display anonymous messages alongside non anonymous ones (2 weeks ago) \<Struan Donald>
* 9a67fd8 - handle anonymous authors in the message display template (2 weeks ago) \<Struan Donald>
* 50c8890 - update message form to validate author name (2 weeks ago) \<Struan Donald>
* 09a2b01 - update message model to allow blank author name (2 weeks ago) \<Struan Donald>
* df25173 - instance config option to allow anonymous user names (2 weeks ago) \<Struan Donald>

PR #1166, opened 2016-07-07
WIP Add a custom_stylesheet template tag to load instance css

struan:
This add a a templatetag which checks for an instance specific
stylesheets in /static/{instance_name}/css/custom.css. If this is
present then it outputs the link tag to load the stylesheet otherwise an
empty string.

### DIFF
--- a/nuntium/templatetags/nuntium_tags.py
+++ b/nuntium/templatetags/nuntium_tags.py
@@ -1,6 +1,9 @@
 import os
+import re
 from django import template
 from django.conf import settings
+from django.core.exceptions import SuspiciousOperation
+from django.utils.translation import ugettext_lazy as _
 from django.contrib.staticfiles import finders
 from contactos.models import Contact
 from django.db.models import Q
@@ -47,6 +50,12 @@ def custom_stylesheet(context):
         subdomain = context['request'].subdomain
     except AttributeError:
         return output
+
+    if not subdomain:
+        return output
+
+    if re.search(r'[^_a-zA-Z0-9\-]', subdomain):
+        raise SuspiciousOperation(_('Invalid subdomain for custom CSS'))
 
     path = os.path.join(subdomain, 'css/custom.css')
     file_path = finders.find(path)
--- a/nuntium/templatetags/nuntium_tags.py
+++ b/nuntium/templatetags/nuntium_tags.py
@@ -43,10 +43,15 @@ def assignment_url_with_subdomain(context, view, subdomain=UNSET, *args, **kwarg
 @register.simple_tag(takes_context=True)
 def custom_stylesheet(context):
     output = ''
-    if hasattr(context['request'], 'subdomain') and context['request'].subdomain is not None:
-        path = os.path.join(context['request'].subdomain, 'css/custom.css')
-        file_path = finders.find(path)
-        if file_path is not None:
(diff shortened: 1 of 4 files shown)


struan:
I'm not totally sure about the location for the files here, but putting them under the static directories of the apps didn't seem like the right place.

nuntium/templatetags/nuntium_tags.py
```diff
@@ -35,3 +38,15 @@ def join_with_commas(obj_list):
 @register.assignment_tag(takes_context=True)
 def assignment_url_with_subdomain(context, view, subdomain=UNSET, *args, **kwargs):
     return subdomainsurls(context, view, subdomain, *args, **kwargs)
+
+
+@register.simple_tag(takes_context=True)
+def custom_stylesheet(context):
+    output = ''
+    if hasattr(context['request'], 'subdomain') and context['request'].subdomain is not None:
```
mhl: I _think_ it's more Pythonic / duck-typing-style to just try to access the attribute rather than test with hasattr. So I might write this as:

```
try:
    subdomain = context['request'].subdomain
except AttributeError:
    return output
```

... and then use `subdomain` rather than `context['request'].subdomain` in the rest.

nuntium/templatetags/nuntium_tags.py
```diff
@@ -35,3 +38,15 @@ def join_with_commas(obj_list):
...
+        path = os.path.join(context['request'].subdomain, 'css/custom.css')
```
mhl: I've tried to convince myself that the subdomain can't be manipulated to contain any characters that might cause problems, but that's relatively tricky for me to do, anyway.  To be defensive, I think I'd put in a check that the subdomain only contain a limited range of characters (`[-_a-zA-Z]` perhaps?) and rather an exception otherwise.

writeit/settings.py
```diff
@@ -88,8 +88,11 @@
 # Example: "http://example.com/static/", "http://static.example.com/"
 STATIC_URL = '/static/'
 
+INSTANCE_STATIC = os.path.abspath(os.path.join(BASE_DIR, '..', 'static'))
+
 # Additional locations of static files
 STATICFILES_DIRS = (
+    INSTANCE_STATIC,
```
mhl: The new `instance` app contains the model that represents a subdomain, so perhaps it would be simpler to say custom CSS for subdomains has to go under: `instance/static/` - then you wouldn't need this change.

nuntium/templatetags/nuntium_tags.py
mhl: I'd have a slight preference for the paths to be like:

    css/custom/australia-representatives.css
    css/custom/kenyan-politicians.css

... rather than:

    australia-representatives/css/custom.css
    kenyan-politicians/css/custom.css

... especially if you switch to suggesting they go under `instance/static`, which might have other CSS there at some point.

struan: My reasoning for this was that if you want to add quite a bit of custom static content - e.g images etc, being able to group it all in one place would make things easier to manage.

writeit/settings.py
struan: yup, that makes some sense.

mhl:
This has been merged as d1a62f5b9afa so I'm closing it

PR #1168, opened 2016-07-08
WIP: Allow the site to be served over https

chrismytton:
A few changes to make WriteInPublic work over https.

Related to #451

### DIFF
--- a/nuntium/templates/base_manager.html
+++ b/nuntium/templates/base_manager.html
@@ -13,7 +13,7 @@
     <meta name="description" content="">
     <meta name="author" content="">
     <link rel="shortcut icon" href="../../docs-assets/ico/favicon.png">
-    <link href='http://fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic' rel='stylesheet' type='text/css'>
+    <link href='https://fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic' rel='stylesheet' type='text/css'>
     {% if WEB_BASED %}
     <title>{{ writeitinstance|default:'WriteInPublic' }}</title>
     {% else %}
--- a/writeit/settings.py
+++ b/writeit/settings.py
@@ -67,6 +67,10 @@ USE_L10N = True
 # If you set this to False, Django will not use timezone-aware datetimes.
 USE_TZ = True
 
+# If the site is being served over https change this to 'https' so that links
+# to subdomains use the correct scheme.
+DEFAULT_URL_SCHEME = 'http'
+
 # Absolute filesystem path to the directory that will hold user-uploaded files.
 # Example: "/var/www/example.com/media/"
 MEDIA_ROOT = ''


nuntium/templates/base_manager.html
```diff
@@ -13,7 +13,7 @@
     <meta name="description" content="">
     <meta name="author" content="">
     <link rel="shortcut icon" href="../../docs-assets/ico/favicon.png">
-    <link href='http://fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic' rel='stylesheet' type='text/css'>
+    <link href='https://fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic' rel='stylesheet' type='text/css'>
```
mhl: Might this be better as a protocol relative URL, i.e.:

```
    <link href='//fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic' rel='stylesheet' type='text/css'>
```

writeit/settings.py
```diff
@@ -67,6 +67,10 @@
 # If you set this to False, Django will not use timezone-aware datetimes.
 USE_TZ = True
 
+# If the site is being served over https change this to 'https' so that links
+# to subdomains use the correct scheme.
+DEFAULT_URL_SCHEME = 'http'
+
```
mhl: I think this is only relevant for the django-subdomains' `urljoin` function, which isn't used in write-it or in the rest of django-subdomains, so I probably wouldn't bother with this.

chrismytton: Looks like it's also used in [django-subdomains `reverse`](https://github.com/tkaemming/django-subdomains/blob/db7f08db7586ba7b4f713aebb83daf2621edde6b/subdomains/utils.py#L62), which we use instead of Django's standard `reverse`. So I think we will need to change this to correctly link to subdomain?

nuntium/templates/base_manager.html
chrismytton: I think _always_ serving from https is now the recommended approach. See [the update at the top of this article](http://www.paulirish.com/2010/the-protocol-relative-url/) for more details.

mhl: OK, fair enough - good to know.

writeit/settings.py
mhl: Oops, silly of me to miss that when grepping... BTW, I see that you already set `DEFAULT_URL_SCHEME = {% if enable_https %}'https'{% else %}'http'{% endif %}` in `local_settings.py` in Ansible, so that's why this was working for me anyway.

mhl:
:+1: Looks good. The test failure I hope should be fixed by: https://github.com/ciudadanointeligente/write-it/pull/1188 - could you check that, possibly? Then we can merge that, rebase this and hopefully see the tests pass.

chrismytton:
@mhl Awesome, I've reviewed #1188 now. Please feel free to take over this branch and rebase/merge it when you're ready :)

PR #1169, opened 2016-07-08
Fix some DateField defaults (should be 'now', not 'now()')

mhl:
The default for two DateField fields was now(), which would evaluate to
the current date and time at the point when the class is defined; what's
usually intended is that the default is the function 'now', which will
be evaluted when an instance of the model is created.  This commit makes
that change, including the corresponding migration.


struan:
Looks fine to me.

PR #1173, opened 2016-07-15
Better celery docs

chrismytton:
- Update the README section about celery worker to remove the reference to PopIt.
- Add a new README section about celery beat, which is needed to send emails to politicians.
- Tell user how to run celery beat in vagrant box

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,13 +37,20 @@ Then run the development server with:
 
 And visit http://127.0.0.1.xip.io:8000 on your host machine to use WriteIt.
 
-You will also need to start a celery worker with:
+### Background jobs
 
     ./manage.py celery worker
 
-which will handle syncing contact details from popit instances. If you
+This handles syncing contact details from remote sources. If you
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,14 +37,21 @@ Then run the development server with:
...
-which will handle syncing contact details from popit instances. If you
+This handles syncing contact details from remote sources. If you
 have created a new instance and the contacts do not seem to be syncing
 it is probably because a celery worker is not running.
 
+### Scheduled jobs
+
+    ./manage.py celery beat
+
+This sends emails to politicians and periodically re-sync contacts from
```
tmtmtmtm: I think 'politicians' is a little too precise here. Probably better as 'recipients' or equivalent.

tmtmtmtm:
One nit in place, otherwise looks good.

README.md
chrismytton: Yep good point, fixed.

PR #1174, opened 2016-08-04
Add extra languages and Transifex configuration

mhl:
These commits add:
* A dependency on the Transifex command-line client, for convenience
* Two new languages (cs_CZ and fa)
* Uses of `./manage.py compilemessages` that are required for multi-lingual support


struan:
Looks good to me.

PR #1177, opened 2016-08-22
allow messages created before moderation enabled to be moderated

### LINKED ISSUE #704: Moderation needed in all messages setting only works for new messages
When turning this setting on existing messages can't be moderated because they don't have a `Moderation` object associated with them.

struan:
Prevent an exception being thrown when moderating a message confirmed
before moderation for all messages by creating the moderation object at
moderation time.

Fixes #704

### DIFF
--- a/nuntium/models.py
+++ b/nuntium/models.py
@@ -303,7 +303,18 @@ class Message(models.Model):
             raise ValidationError(_('The message needs '
                 + 'to be confirmated first'))
         self.set_to_ready()
-        self.moderation.success()
+        # if we turn on moderation after some messages have been created then
+        # they will not have associated Moderation objects so we need to catch
+        # that, create the moderation object and then try again.
+        try:
+            self.moderation.success()
+        except Moderation.DoesNotExist:
+            if self.writeitinstance.config.moderation_needed_in_all_messages:
+                moderation, created = Moderation.objects.create(message=self)
+                if created:
+                    self.moderation.success()
+                    return
+            raise
 
     def __unicode__(self):
         return _('%(subject)s at %(instance)s') % {
--- a/nuntium/tests/moderation_messages_test.py
+++ b/nuntium/tests/moderation_messages_test.py
@@ -212,6 +212,53 @@ A gaf i dynnu sylw'r Prif Weinidog hefyd at y sefyllfa yn Ysbyty Gwynedd yn ddiw
         outbound_message_to_pedro = OutboundMessage.objects.get(message=message)
         self.assertEquals(outbound_message_to_pedro.status, 'new')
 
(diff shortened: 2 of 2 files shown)


nuntium/models.py
```diff
@@ -303,7 +303,17 @@ def moderate(self):
...
         self.set_to_ready()
-        self.moderation.success()
+        # if we turn on moderation after some messages have been created then
+        # they will not have associated Moderation objects so we need to catch
+        # that, create the moderation object and then try again.
+        try:
+            self.moderation.success()
+        except Moderation.DoesNotExist:
+            if self.writeitinstance.config.moderation_needed_in_all_messages:
+                moderation, created = Moderation.objects.get_or_create(message=self)
```
mhl: We know the moderation object doesn't exist, so this should just be `Moderation.objects.create()` so people don't wonder about how it might actually exist :)

nuntium/models.py
```diff
@@ -303,7 +303,17 @@ def moderate(self):
...
+                if created:
+                    return self.moderation.success()
```
mhl: Since the `return` here is just for control flow rather than explicitly returning a `None`, I'd split it into:
```
self.moderation.success()
return
```

nuntium/tests/moderation_messages_test.py
```diff
@@ -212,6 +212,53 @@ def test_message_that_has_not_been_confirmed_cannot_be_moderated(self):
...
+        with self.assertRaises(Moderation.DoesNotExist):
+            message.moderate()
+
+        self.assertFalse(message.moderated)
+        outbound_message_to_pedro = OutboundMessage.objects.get(message=message)
+        self.assertEquals(outbound_message_to_pedro.status, 'ready')
+
+    # for https://github.com/ciudadanointeligente/write-it/issues/704
+    def test_moderation_created_for_old_messages_after_moderation_turned_on(self):
+        # this message has not been confirmed
```
mhl: "has not been confirmed" seems misleading here, since in the next statement you set it to be confirmed - maybe I'd add a "now set it as confirmed" comment afterwards.

mhl:
> Prevent an exception being thrown when moderating a message confirmed
before moderation for all messages by creating the moderation object at
moderation time.

Could this commit message be a bit clearer? (I think the way it's expressed in the issue is easier to understand.) Maybe just leaving out "for all messages"?

mhl:
:+1:  Looks good - I left a few comments, but they're all very minor things.

mhl:
In @struan's absence I've squashed those additional commits and slightly edited the commit message - I hope that's OK. (The alpaca branch can still be rewritten without much consequence, if not, of course.)

PR #1178, opened 2016-08-24
move moderation to web based queue

struan:
This PR turns off the email based moderation and instead adds a page in the `/manage/` section of an instance with a queue of messages to be moderated.

Moderation of messages then takes place on the message detail page with the addition of a Reject button along side the existing Accept button.

This also fixes some existing issues:
 - If moderation is off then all public messages are marked as moderated upon confirmation
 - An error is thrown if a previously moderated message is moderated again
 - Prevents moderation being turned off if there are messages in the moderation queue


mhl:
The biggest issue for me here, before really considering the content, is that this turns off the email based moderation completely. I'd been assuming that we're planning to merge the `alpaca` branch into `master` as soon as we're happy with it, in which case we wouldn't want to turn off email moderation for every existing instance. Could there be a feature flag per instance for this?

nuntium/templates/nuntium/profiles/moderation_queue.html
```diff
@@ -24,7 +24,7 @@
     <a href="{% url 'writeitinstance_answernotification_update' subdomain=writeitinstance.slug %}" class="page-header__settings-link btn btn-default btn-sm"><i class="glyphicon glyphicon-wrench"></i> {% trans "Settings" %}</a>
   </div>
   <div class="manager-overview">
-    {% blocktrans count message_count=writeitinstance.message_set.count %}
+    {% blocktrans count message_count=message_list.count %}
```
mhl: This fixup should be squashed

nuntium/user_section/views.py
```diff
@@ -484,16 +484,7 @@ class ModerationQueue(LoginRequiredMixin, ListView):
...
-            .get_queryset() \
-            .filter(writeitinstance=self.writeitinstance) \
-            .exclude(moderated=True) \
-            .exclude(confirmated=False)
-
-        if not self.writeitinstance.config.moderation_needed_in_all_messages:
-            qs = qs.exclude(public=True)
-
-        return qs
+        return Message.moderation_required_objects.filter(writeitinstance=self.writeitinstance)
```
mhl: I think this is a fixup to the first commit too

nuntium/models.py
```diff
@@ -96,8 +96,8 @@ def get_queryset(self):
 class NonModeratedMessagesManager(MessagesManager):
     def get_queryset(self):
         queryset = super(NonModeratedMessagesManager, self).get_queryset()
-        return queryset.filter(Q(public=True), Q(confirmated=True))\
-            .exclude(Q(moderated=True) | Q(moderated=None))
+        return queryset.filter(Q(confirmated=True))\
```
mhl: The `Q` is unnecessary here

mhl:
I've rebased this now - I had to resolve a conflict between 35f547758897b8027a0d1 and 88e3c97cb4fd

PR #1180, opened 2016-09-06
Add gettext dependency to vagrant provision.sh

zarino:
Not sure whether this has been added in the right place, but I needed gettext to make django work in the vagrant vm.

I've also swapped the order of the setup steps in the shell login message, because if you don't load the example data first, django will automatically redirect you to example.com and you'll have no idea why.


zarino:
@mhl – done!

PR #1181, opened 2016-09-06
WIP: Migrate from popit-django to django-popolo

mhl:
[Note: this is not completely finished - there are some further things I would like to address, which I've noted below.]

PopIt is being deprecated, so it no longer makes sense for write-it / WriteInPublic to depend on PopIt instances as sources of data, or for it to depend on the deprecated PopIt client, `popit-django`. It makes more sense now to move to a model where data sources are all URLs of Popolo JSON, and to use the `django-popolo` project to store the data from such Popolo JSON sources.

This pull request migrates from using `popit-django` and PopIt to using `django-popolo` and Popolo JSON URLs instead.

### Notes for reviewers / self review

I've opted to keep these as a large number of quite small commits, since I think that will make it much easier to review: some commits are quite mechanical (more or less a global search-and-replace) and you would want them separate from changes that are more involved, for example.

I believe this works, but there are some changes that I think should still be made, or things I'm concerned about:

1. `instance/models.py` has code that parses Popolo JSON and updates `django-popolo` models, but there is already code that essentially does the same thing in django-popolo's `popolo/importers/popit.py` - it would be better to reuse that, although we'd need to make some changes, e.g.:
    * Add callbacks for when an object is created, so we can add custom identfiers.
    * Allow you to customize (e.g. with an overridable method) how an existing model for updating is detected
   * Make `django-popolo`'s importer delete objects that have disappeared.
   * Make sure long names are truncated to fit the `django-popolo` models' fields' maximum lengths.
* (This could be dealt with by the former point, but...) At the moment `popolo.Organization` objects from different `PopoloSource`s aren't distinguished - if they happen to have the same ID, they'll be confused.
* The issue that names may need to be truncated on import applies to this importer too.
* The import of all the data currently takes place in a single transaction, which might take a very long time; this has various risks, depending on the underlying database (e.g. "database locked" errors, or very slow updates). This transaction should be unneccessary, since the models being updated are only by this code, otherwise they're just read from. We would need to ensure that a manual update triggered from the interface can't take place at the same time as the weekly run, however.
* `InstanceMembership` objects are deleted and recreated on resyncing from the popolo source. We need to double-check if the IDs of these objects are preserved across steps of writing to a politician, or else just change the code to make sure the minimal updates are done instead of deleting all and recreating.
* Though the tests pass, more hand testing of the site with this new code would be worthwhile; e.g. now-renamed attribute names in templates wouldn't generate errors, but could cause odd missing text on the site.


instance/migrations/0004_add_django_popolo_people.py
```diff
@@ -0,0 +1,190 @@
...
+        split_as_list[2] = '/api/v0.1/export.json'
+        split_as_list[3] = ''
+        split_as_list[4] = ''
+        url = urlunsplit(split_as_list)
+    else:
+        url = original_url
+    return url
+
+
+def forwards(apps, schema_editor):
```
lfalvarez: woow this is incredibly cautious particularly in preserving the old data!
I get the impression that this is a pretty interesting piece of code and we should be aware of what comes out of processing this migration. Anyhow, :+1:

mhl: :smile: The scariest bit of this migration for me (although this whole PR is quite scary!) is actually the `update_source_url`, function above, since there are so many different PopIt URLs in the live database, lots of which don't work at the moment anyway. I hope that `update_source_url` will produce working URLs for those were working previously, but I think it's particularly worth double-checking that.

instance/models.py
```diff
@@ -1,68 +1,317 @@
...
 from subdomains.utils import reverse
 
+from contactos.models import Contact
+from mailit import MailChannel
+
+
+class PopoloPersonQuerySet(models.QuerySet):
+
+    def get_from_api_uri(self, uri_from_api):
+        try:
```
lfalvarez: Abosolutely understood :+1:

mhl: I've made the fix that you suggested in this fixup: d579dabc0e70c5490d712b50823e8e82fdc150d8

PR #1183, opened 2016-09-13
hide messages that have not been moderated

struan:
If you directly access a message's page then hide the message if it has
not been moderated.


struan:
I don't understand the test failures on postgres :(

chrismytton:
Looks good! 👍  🚀

PR #1189, opened 2016-09-20
Hide unmoderated messages on from page

struan:
Make sure the we aren't displaying unmoderated message on the "all messages from this person".

Also add a test to the "all messages to" tests to check for this as well.

Fixes mysociety/alpaca#24


mhl:
:+1: Looks good, thanks!

PR #1191, opened 2016-09-21
More moderated viewing fixes

struan:
Fixes the logic for the /read page so we do not hide messages that have been created before the current moderation code and have moderated=None.

Also tidy up the other message views to use the public_objects manager of Message rather than repeating the same filters in multiple place.

### DIFF
--- a/nuntium/tests/moderation_messages_test.py
+++ b/nuntium/tests/moderation_messages_test.py
@@ -74,6 +74,51 @@ class ModerationMessagesTestCase(TestCase):
         self.assertEquals(response.context['awaiting_moderation'], True)
         self.assertContains(response, 'This message is awaiting moderation')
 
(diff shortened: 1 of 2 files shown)


nuntium/tests/moderation_messages_test.py
```diff
@@ -74,6 +74,55 @@ def test_cannot_view_unmoderated_message(self):
         self.assertEquals(response.context['awaiting_moderation'], True)
         self.assertContains(response, 'This message is awaiting moderation')
 
+    def test_can_view_moderated_message(self):
+        self.writeitinstance1.config.save()
```
mhl: Is this line (and the similar one below) actually needed?

struan: oh, no, good point.

PR #1195, opened 2016-09-23
Migration to django-popolo part 2

mhl:
This is the second part of: #1181 for review. See that pull request for the context for this pull request.

This removes the old popit-django model references.

### DIFF
--- a/contactos/models.py
+++ b/contactos/models.py
@@ -21,7 +21,7 @@ class ContactType(models.Model):
 class Contact(models.Model):
     """docstring for Contact"""
     contact_type = models.ForeignKey('ContactType')
-    popolo_person = models.ForeignKey(Person)
+    person = models.ForeignKey(Person)
     value = models.CharField(max_length=512)
     is_bounced = models.BooleanField(default=False)
     owner = models.ForeignKey(User, related_name="contacts", null=True)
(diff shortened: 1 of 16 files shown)


nuntium/migrations/0003_denull_popolo_people.py
```diff
@@ -0,0 +1,41 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='answer',
+            name='popolo_person',
+            field=models.ForeignKey(to='popolo.Person'),
+            preserve_default=True,
+        ),
+        migrations.AlterField(
```
mhl: Oops, this `AlterField` and the next one are spurious - they were accidentally generated by having the old models file present after 1a0c81a4e78 and running makemigrations, I'll add a fixup to remove them.

mhl: And actually the same spurious AlterFields got into: b6fbffba8104

contactos/models.py
```diff
@@ -22,8 +21,7 @@ def __unicode__(self):
 class Contact(models.Model):
     """docstring for Contact"""
     contact_type = models.ForeignKey('ContactType')
-    person = models.ForeignKey(Person)
-    popolo_person = models.ForeignKey(PopoloPerson, null=True, blank=True)
+    person = models.ForeignKey(PopoloPerson)
```
chrismytton: I think we can import `PopoloPerson` as `Person` now that we've removed the reference to the `popit.models.Person`?

mhl: Good point, I'll do that.

PR #1198, opened 2016-09-27
Handle people who are missing email addresses more elegantly

mhl:
After this pull request, people without email addresses are also offered as possible selections to write to. They are clearly marked as such, and users are given the option of either removing that recipient, or going to a page which lists all such people and appeals to the user to help in finding those email addresses. Here's @zarino's nice animation of the new selection process:

dropdown

Things still to do:
- [ ] Add a "send email to the instance owner" contact form.
- [x] Look at an issue with the layout of the drop-down's text in RTL languages.

### DIFF
--- a/nuntium/static/js/who.js
+++ b/nuntium/static/js/who.js
@@ -20,7 +20,8 @@ $(function(){
     var choicesSentence = interpolate(
       gettext("%(open_danger_tag)sRemove this recipient%(close_tag)s or %(open_default_tag)sContribute new contact details%(close_tag)s"),
       {open_danger_tag: '<a class="btn btn-danger btn-sm js-deselect">',
-       open_default_tag: '<a class="btn btn-default btn-sm" href="http://example.com" target="_blank">',
+       open_default_tag: '<a class="btn btn-default btn-sm" href="' +
+       missingContactsURL + '" target="_blank">',
        close_tag: '</a>'},
       true);
     $warning.attr('role', 'alert');
--- a/nuntium/static/js/who.js
+++ b/nuntium/static/js/who.js
@@ -1,5 +1,26 @@
 $(function(){
 
+  // Temporarily add the "uncontactable" class to a random selection
+  // of contacts in the person selector.
+  // TODO: This should be done in forms.py > PersonMultipleChoiceField
+  $('.chosen-person-select option').each(function(i){
+    if( Math.random(0, 1) <= 0.2 ){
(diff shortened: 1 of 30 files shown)


nuntium/static/js/who.js
```diff
@@ -0,0 +1,87 @@
...
+      if(originalOptElement.classes.indexOf('uncontactable') !== -1){
+        $(this).addClass('uncontactable');
+      }
+    });
+  });
+
+  var showUncontactableWarning = function showUncontactableWarning($selectedOption, $originalSelect, $chosenSelect){
+    var $warning = $('<div>').addClass('alert alert-danger uncontactable-alert');
+    var warningSentence = interpolate(
+      gettext("We do not have contact details for %s."),
```
zarino: We use “do not” here, but “don’t” on the `missing_contacts.html` page. Maybe we should standardise? How conversational is the rest of the WiP UI?

nuntium/templates/write/missing_contacts.html
```diff
@@ -0,0 +1,32 @@
...
+{% endblock %}
+
+
+{% block content_inner %}
+
+  <h1>{% trans "Help to find missing contact details" context "Title of page appealing for help finding more contact details" %}</h1>
+
+  <p>
+    {% blocktrans trimmed %}
+      We don't have an email address for any of the people
```
zarino: Tiny thing, but, curly quote here? `don’t`, rather than `don't` 😄 

(We use curly quotes elsewhere, like `"You can’t currently create new messages using this site."` in `who.html`)

writeit/templates/subdomains/iran/write/missing_contacts.html
```diff
@@ -0,0 +1,33 @@
...
+{% block content_inner %}
+
+    <div class="full-width-wrapper write-wrapper">
+      <main class="centred write">
+
+        <h1>{% trans "Help to find missing contact details" context "Title of page appealing for help finding more contact details" %}</h1>
+
+        <p>
+          {% blocktrans trimmed with email='support@namehbeanha.com' %}
+            We don't have an email address for any of the people
```
zarino: Same comment about curly quotes, as in `nuntium/templates/write/missing_contacts.html`

nuntium/templates/write/missing_contacts.html
mhl: heh, OK, i'll fix this and the other one

Issue #1194, opened 2016-09-23
Allow grouping of messages based on email and name


struan:
At the moment the `/from/` page only uses email on to group messages from the same person on one page.

In some circumstances it may be useful to group based on name rather than email to avoid it being obvious that a set of messages were all sent by the same person.

We should provide an option to group by name *and* email for such circumstances. (We need both to stop people with the same name but different emails being grouped together)

This option should be off by default, and possibly not exposed in the admin interface as in most cases grouping by email is the correct thing to do.

PR #1201, opened 2016-11-15
Add an instance config option for stricter anonymity

mhl:
There is a view ('all-messages-from-the-same-author-as') which, given a
message slug, allows you to list any other messages with the same author
email address as the message with that slug. In situations where anonymity
is important, this is undesirable - someone might send a sensitive message
with a false name and an uncontroversial one with their real name, in
which case this view would reveal that they're sent by the same person.

This commit introduces a new instance config option
('email_and_name_must_match') which means that the
'all-messages-from-the-same-author-as' view will only show messages that
have the same author_email *and* author_name, to avoid this information
leak.

Fixes mysociety/alpaca#35
Fixes ciudadanointeligente/write-it#1194

### DIFF
--- a/instance/migrations/0003_add_name_and_email_match_config_option.py
+++ b/instance/migrations/0003_add_name_and_email_match_config_option.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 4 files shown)


struan:
This looks good to me.

PR #1209, opened 2017-01-04
An 'updated' DateTimeField should use auto_now, not auto_now_add

mhl:
auto_now_add=True only sets the date when the object is first created,
which is only suitable for the 'created' field. auto_now=True is the
correct option for 'updated'.

This pull request also contains a commit to make sure that the
'inprogress' status is set however syncing is initiated, and one to remove
some unnecessary code.

### DIFF
--- a/instance/models.py
+++ b/instance/models.py
@@ -311,11 +311,12 @@ class WriteItInstance(models.Model):
         pass
 
     def _load_persons_from_popolo_json(self, popolo_source):
-        success_relating_people, error = self.relate_with_persons_from_popolo_json(popolo_source)
         record, created = WriteitInstancePopitInstanceRecord.objects.get_or_create(
             writeitinstance=self,
             popolo_source=popolo_source
             )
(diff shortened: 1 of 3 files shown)


instance/models.py
```diff
@@ -311,11 +311,12 @@ def do_something_with_a_vanished_popit_api_instance(self, popolo_source):
         pass
 
     def _load_persons_from_popolo_json(self, popolo_source):
-        success_relating_people, error = self.relate_with_persons_from_popolo_json(popolo_source)
         record, created = WriteitInstancePopitInstanceRecord.objects.get_or_create(
             writeitinstance=self,
             popolo_source=popolo_source
             )
+        record.set_status('inprogress')
+        success_relating_people, error = self.relate_with_persons_from_popolo_json(popolo_source)
```
struan: Might be nice to see an explanation in the commit message of why the order of this has been moved.

mhl: Done.

PR #1210, opened 2017-01-05
raise and handle exception on temporary mail failure

struan:
If we get a temporary failure message then ignore it as we only need to
do something if the failure is permanent.

### DIFF
--- a/mailit/bin/handleemail.py
+++ b/mailit/bin/handleemail.py
@@ -12,7 +12,7 @@ from flufl.bounce import all_failures, scan_message
 from mailit.models import RawIncomingEmail
 from nuntium.models import Answer
 from mailit.bin.froide_email_utils import FroideEmailParser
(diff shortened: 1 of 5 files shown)


mailit/bin/handleemail.py
```diff
@@ -152,7 +152,9 @@ def instanciate_answer(self, lines):
         msg = email.message_from_string(msgtxt)
         temporary, permanent = all_failures(msg)
 
-        if temporary or permanent:
+        if temporary:
```
mhl: I'm not sure if in practice this happens with Exim, but `flufl` might return both email addresses with temporary failures and permanent failures, and if there are any in `permanent` it should still register a bounce. So maybe this condition should be `if temporary and not permanent`. (Or switch the cases round so it's `if permanent ... elif temporary ... else ... end`.)

PR #1212, opened 2017-01-09
Fixes for multiple popolo_uri Identifier objects

mhl:
We were getting duplicate identifiers on each resync with scheme='popolo_uri'.
This pull request fixes that.

Fixes mysociety/alpaca#60

### DIFF
--- a/instance/management/commands/instance_remove_extraneous_popolo_uri.py
+++ b/instance/management/commands/instance_remove_extraneous_popolo_uri.py
@@ -1,0 +1,56 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from instance.models import PopoloPerson
+
+BAD_SUBSTRING = '#person-person/'
+
+
+class Command(BaseCommand):
+
+    help = "Remove any extraneous Identifier objects with scheme 'popolo_uri'"
+
+    @transaction.atomic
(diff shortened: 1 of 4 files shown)


instance/management/commands/instance_remove_extraneous_popolo_uri.py
```diff
@@ -0,0 +1,56 @@
...
+                        person_id=person.id,
+                        bad_identifier=sole_identifier.identifier)
+                    errors_found = True
+                    continue
+            # Otherwise we have more than one identifier. Delete any
+            # of the bad legacy IDs:
+            qs.filter(identifier__contains=BAD_SUBSTRING).delete()
+            # Now all the remaining identifiers should be unique. If
+            # they're not, that's an error.
+            unique_remaining = set(qs.values_list('identifier', flat=True))
```
struan: This comment doesn't really match up with the code that follows in that the code errors if there is more than one identifier but the comment reads as if it's only a problem if there are duplicated identifiers. Or at least it does to me :)

mhl: Apparently I forgot what the word "unique" means when writing that. I've changed it to say "should be the same" instead of "should be unique", which I think makes sense!

PR #1214, opened 2017-02-01
[WIP] changes for a Django 1.8 upgrade

mhl:
This is incomplete, and I'm pushing it for reference in case it's of
use - the largest part missing, if I remember right, was getting the
Celery upgrade working.


setup.py
```diff
@@ -15,7 +15,6 @@
 'pyelasticsearch',
 'celery',
 'pyyaml',
-'django-celery',
```
dracos: If it's still in requirements/settings, should it not be here too?

mhl: Oh yes, good point! (That's because at first I took out django-celery completely, and then realised it would be helpful to still have for the results backend...)

